PR #5, opened 2022-06-28
Feature/wfnews 265

ssylver93:


dhlevi (approved):
Good enough for the initial commit. We will want to address some of the dependencies at a later point but this will get our initial component stubs in place.

Note that the build actions aren't in place yet until we establish the AWS container stuff, so actions will currently be failing until that's addressed.

PR #9, opened 2022-06-29
Feature/wfnews 272

ssylver93:


dhlevi (changes requested):
Spring needs an update to the latest. Match up with the API
Worth doing a dependency check and removing anything we dont need or aren't using

dhlevi (approved):

PR #11, opened 2022-07-14
adapt new UI common components 

yzlucas:


dhlevi (changes requested):
There are a few squaks about spring-security-core, may as well give that a version bump and verify the tomcat build.

Other code changes look good for the component inclusion

PR #12, opened 2022-07-14
Feature/wfnews 255

ssylver93:


dhlevi (approved):
Code changes look fine. My only concern is removing the effects for the map if we want to use that to maintain state, but we can always pull them back in from the commit history

PR #13, opened 2022-07-21
Feature/wfnews 341

ssylver93:


dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/wf-admin-dashboard-container/wf-admin-dashboard-container.component.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <p class="dashboard-title">Dashboard: 2022/2023</p>
+    <p class ="active-incidents"><b>84 ACTIVE</b> Fire Incidents</p>
```
dhlevi, line 3: I'm assuming the dates and Numbers is intended to be dynamic once we have values coming in from the API?

PR #18, opened 2022-07-27
Linting auto-fix

yzlucas:
This is the linter auto-fix. 
However I find it changes 'let' to 'const' in many places by preference, not sure if this is what we want


dhlevi:
@yzlucas I've merged the previous PR, which has caused a lot of merge conflicts from this one. Looks like it was using a branch that was significantly out of date. You may want to close this PR, update your local, and try again

PR #20, opened 2022-07-28
restructure active wildfire map

qqnluaq:
wfnews-314, 275, 330, 276,


ssylver93 (approved):

yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -0,0 +1,207 @@
...
+    ) {
+        this.incidentsServiceUrl = this.appConfig.getConfig().rest['newsLocal']
+        // console.log(this.incidentsServiceUrl)
+    }
+
+    ngOnInit() {
+
+        this.appConfig.configEmitter.subscribe((config) => {
+            let mapConfig = []
+            // this.checkMapServiceStatus()
```
dhlevi, line 54: If we don't actually need the commented out code, we should remove it

client/wfnews-war/src/main/angular/src/app/components/base/base.component.ts
```diff
@@ -101,9 +101,9 @@ export class BaseComponent implements OnInit, OnChanges, AfterViewInit {
     }
 
     initializeConnectionService() {
-        this.connectionService.monitor().subscribe(isConnected => {
-            this.isConnected = isConnected.hasNetworkConnection;
-        });
+        // this.connectionService.monitor().subscribe(isConnected => {
```
dhlevi, line 104: If we don't need it, remove it

client/wfnews-war/src/main/angular/src/app/components/wf-map-container/wf-map-container.component.scss
```diff
@@ -68,83 +84,83 @@
...
-                    align-items: center;
-
-                    i.material-icons {
-                        font-size: 20px;
-                    }
-                }
-            }
-        }
-    }
+    // .smk-bespoke-panel .smk-body .smk-bespoke-content {
```
dhlevi, line 87: If we don't need it, remove it

client/wfnews-war/src/main/angular/src/app/services/map-config.service/layers/index.ts
```diff
@@ -1,34 +1,83 @@
...
+		...FirePerimetersLayerConfig( ls ),
+		// ...SmokeForecastLayerConfig( ls ),
+		...PrescribedFireLayerConfig( ls ),
+		// ...WeatherLayerConfig( ls ),
+        // ...WeatherStationsLayerConfig( ls ),
+        // ...PrecipitationLayerConfig( ls )
+	]
 }
+
+// export function notificationDetailLayerConfig( res: AppResourcesConfig ) {
```
dhlevi, line 50: If we don't need it, remove it

dhlevi (approved):
There's quite a bit of commented out blocks. If we don't need them, we should remove them so we don't have clutter in the code. If they're needed later, just add a comment indicating why it's being left in

Otherwise, changes look good. We can address cleanup in a later PR due to the volume of changes here.

PR #21, opened 2022-07-29
adapt wfcc updated header - "news admin header"

yzlucas:


dhlevi:
@yzlucas Recent changes look like they've resulted in some conflicts that will need to be addressed.

Also noticing a lot of old commits on the history. Make sure you're updating your fork (sync from upstream) and branching from that , we shouldn't see commit messages from a month ago on a newer PR

PR #35, opened 2022-08-05
AdminTable, Dashboard Header, Search and pagination

yzlucas:


ssylver93 (approved):

dhlevi:
Just another reminder about the excess commit history
Also, your configuration for IM is in place in DEV configurations, but not in deploy configurations.
You must ensure that the WAR build functions as well. the config.jsp needs to return data from the app or static properties as well, or the calls will not function when deployed.
Also, when deployed to AWS, we will need to hit integration webade/apis

dhlevi (changes requested):
A few issues:
- Performance: loading the table is very slow for a paged request. We should profile if this is an API issue or something in the store
- Searches are also very slow
- Clear should reset the firecentre selector
- Paging component is incorrect. It says it's showing 5 entries, but is displaying 14 (configuration)
- paging is fast? why is this handled differently?
- CSS needs some work. I get extra scrollbars and some weird behaviour on smaller screens

PR #45, opened 2022-08-26
Non compliance cleanup

dhlevi:
Cleanup for the Non compliance tickets:

WFNEWS-403, 401, 398, 397

Additionally, moved AGOL queries into a service class

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
+++ b/client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
@@ -2,7 +2,7 @@
     <div class="title">
         <h1>Active Wildfires Map</h1>
         <h2 class="wildfire-count">
-            <wf-icon iconName="local_fire_department" iconSize="small"></wf-icon>
+            <wf-icon class="fire-icon" iconName="local_fire_department" iconSize="small"></wf-icon>
             <!-- <i class="material-icons-outlined">local_fire_department</i> -->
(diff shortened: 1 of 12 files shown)


yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
```diff
@@ -2,7 +2,7 @@
     <div class="title">
         <h1>Active Wildfires Map</h1>
         <h2 class="wildfire-count">
-            <wf-icon iconName="local_fire_department" iconSize="small"></wf-icon>
+            <wf-icon class="fire-icon" iconName="local_fire_department" iconSize="small"></wf-icon>
```
ssylver93, line 5: Is this the right icon? It looks different than the one screenshotted in the WFNEWS-397 ticket. I had tried the span class of "material-icons-outlined" when looking at this ticket before and it had been showing up as plain English however.

ssylver93 (commented):
Just one comment on the local_fire_department icon

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
dhlevi: It's the right name.... but it definitely looks different.

I want to look into the icon library we're using and see if maybe it's out of date with the latest/greatest from google

ssylver93 (approved):

PR #46, opened 2022-08-29
Add/Remove custom evac order/alerts

dhlevi:


ssylver93 (approved):

dhlevi:
FYI, I merged in a couple additional remediation tickets. They were supposed to go into seperate branches but apparently I was push-happy

client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/admin-incident-form.component.scss
```diff
@@ -125,6 +125,24 @@
 			padding: .5rem 1rem;
 		}
 
+    .size-quarter {
+			flex: 0 0 25%;
+			box-sizing: border-box;
+			padding: .5rem 1rem;
```
yzlucas, line 131: I usually add paddings in "px". we need to agree on either using rem or px in the future styling

dhlevi: Well, if you prefer an "either or" approach we should only use rem, as it's not an absolute unit like px. However I would argue there are times when we'd benefit from using both

PR #58, opened 2022-08-31
In and out animation for left menu

gpineiro-vivid:


client/wfnews-war/src/main/angular/src/styles/_base.scss
```diff
@@ -533,4 +533,14 @@ iframe#login-iframe {
     #map {
         min-height: $wf1-app-map-mobile-height;
     }
+}
+
+.details {
+    transition: all .2s linear;
+    width: 600px;
```
dhlevi, line 540: Not an issue right now, however we might want to make the width responsive to screen sizes in the future, so might need some media tagging or a max-width

dhlevi (approved):

PR #66, opened 2022-09-01
WFNEWS-260 remaining part

yzlucas:


dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -188,10 +193,39 @@ export class ActiveWildfireMapComponent implements OnInit, OnChanges {
     }
 
     useMyCurrentLocation(){
```
dhlevi, line 195:
```
useMyCurrentLocation(){
        this.searchText = undefined;
        this.commonUtilityService.getCurrentLocationPromise();

        const long = (this.commonUtilityService.getCurrentLocationPromise()['__zone_symbol__value'].coords.longitude);
        const lat = (this.commonUtilityService.getCurrentLocationPromise()['__zone_symbol__value'].coords.latitude);
        if( lat && long ){
            this.showAreaHighlight([long,lat],50)
            this.showLocationMarker({
                type: 'Point',
                coordinates: [long, lat]
            });
        }
    }
```
Note sure why the self ref is there at all? And when appropriate, use `const`. There's also the first call the `getCurrentLocationPromise`... not sure if this is needed, as nothing is done with the call, and it's called again twice right after

dhlevi (approved):

PR #110, opened 2022-09-09
WFNEWS-356: Adding PDF Map upload/table

dhlevi:
Basic components in place, might need some styling adjustment for the default message popup dialog and the table widths, but functionality is in place and ready for review


dhlevi:
FYI, sonar failures are due to the addition of meta @types tags: http://resources.wfdm.nrs.gov.bc.ca/fileMetadataResource
@vivid-cpreston We can add an ignore for these in sonar

PR #111, opened 2022-09-09
add log out option in admin pag, fix the support link in admin header, and WFNEWS-450

yzlucas:


client/wfnews-war/src/main/angular/src/app/app.module.ts
```diff
@@ -82,6 +82,8 @@ import { PanelSmokeForecastComponent } from './components/panel-smoke-forecast/p
 import { PanelAreaRestrictionsComponent } from './components/panel-area-restrictions/panel-area-restrictions.component';
 import { PanelFireDangerComponent } from './components/panel-fire-danger/panel-fire-danger.component';
 import { CommonUtilityService } from "./services/common-utility.service";
+import { CookieService } from 'ngx-cookie-service';
```
dhlevi, line 85: If we're not using this in the code, you can remove it from AppModule. I didn't see any other references to it

dhlevi (changes requested):

dhlevi (approved):

PR #117, opened 2022-09-13
Wfnews-455, add supportLink property into config

yzlucas:


yzlucas:
This is just a tab-group set-up and Active Wildfire List still need to be implemented. The activeWildfireList endpoint in our news-api will need pagination,

dhlevi (approved):

PR #121, opened 2022-09-13
Feature/wfnews 375

ssylver93:


dhlevi:
@ssylver93 Looks like the structure is in place. Do you have unit tests to add?

PR #143, opened 2022-09-16
non-compliance: Wfnews-489, Wfnews-491

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -97,6 +97,7 @@ export class ActiveWildfireMapComponent implements OnInit {
                     this.mapConfig = [...mapConfig, deviceConfig, 'theme=wf', '?'];
                 });
         });
+        this.commonUtilityService.getCurrentLocationPromise()
```
dhlevi, line 100: Minor, but here you're fetching the promise and not resolving it. If the function fails, it may throw an unhandled exception. If needed, make sure to add a `then` and/or `catch` so the promise is resolved cleanly. 

In this case, if the promise doesn't resolve, SMK/leaflet didnt init so we have bigger problems, hence it being a bit of a low priority, more of a "best practice" consideration for the future.

dhlevi (approved):

PR #158, opened 2022-09-17
WFNEWS-291: Add Wildfire Resources static content

dhlevi:


dhlevi:
The duplications noted by sonar are in the static content (use of the same labels). Can be ignored, or fixed by moving the content into a config object.

client/wfnews-war/src/main/angular/src/app/components/wildfire-resources/wf-resources.component.ts
```diff
@@ -0,0 +1,102 @@
...
+  styleUrls: [ './wf-resources.component.scss' ]
+})
+export class WildfirewResourcesComponent {
+  // This can be moved into a config, for easier changes later
+  public resourcePanels: ResourcePanel[] = [
+    {
+      title: 'Preparedness and Prevention',
+      resources: [
+        {
+          title: 'Wildfire Prevention', subtitle: 'BC Wildfire Services', url: 'https://www2.gov.bc.ca/gov/content/safety/wildfire-status/prevention',
```
yzlucas, line 16: mock up's subtitle labeled as "BC Wildfires Service" .
To be consistent, should be "BC Wildfires Service", not plural ?

yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/wildfire-resources/wf-resources.component.ts
dhlevi: Fixed

PR #161, opened 2022-09-19
removed placeholder from incident details form

gpineiro-vivid:


ssylver93 (changes requested):
For the geopgraphicDescription field, the value is showing as "location" rather than "Location" with a capital L in the dev environment. Looks good otherwise!

dhlevi (approved):

PR #164, opened 2022-09-19
checkToken update

yzlucas:
I have touched base with Chris P about secret placeholder {webade-oauth2.wfim-incident-manager-ui.client.secret}

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,4 @@
 {
-  "java.configuration.updateBuildConfiguration": "interactive",
+  "java.configuration.updateBuildConfiguration": "automatic",
   "files.maxMemoryForLargeFilesMB": 8192
 }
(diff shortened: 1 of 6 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,4 @@
 {
-  "java.configuration.updateBuildConfiguration": "interactive",
+  "java.configuration.updateBuildConfiguration": "automatic",
```
dhlevi, line 2: Generally, we don't need to commit settings.json (not everyone will be using vscode). If it requires specific configs, we can document them in the readme.

client/wfnews-war/src/main/resources/static.properties
```diff
@@ -11,3 +11,8 @@ wfim-rest.url=https://i1bcwsapi.nrs.gov.bc.ca/wfim-incidents-api
 siteminder.url.prefix=https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?returl=
 bcWildFireSupportPage=https://intranet.gov.bc.ca/bcws/provincial-programs/strategic-initiatives-and-innovation/wildfire-one/wildfire-one-training
 check.token.url=/checkToken.jsp
+webade-oauth2.client.id=WFIM_INCIDENT_MANAGER_UI
+webade-oauth2.wfim-incident-manager-ui.client.secret=${webade-oauth2.wfim-incident-manager-ui.client.secret}
+webade-oauth2.check.token.v2.url=https://intapps.nrs.gov.bc.ca/pub/oauth2/v1/check_token?disableDeveloperFilter=true
```
dhlevi, line 16: Just a reminder that like the secret token above, we'll have to tokenize all of the configs for environments

dhlevi (approved):

PR #175, opened 2022-09-20
WFNEWS-476 removed legends from incident form sections

gpineiro-vivid:


dhlevi:
This will result in accessibility non-compliance issues. If this is acceptable we'll just have to flag them in sonar as "wont fix", and make sure we have appropriate aria tagging for screen readers.

PR #190, opened 2022-09-22
We may missing Check Token Servlet

yzlucas:


client/wfnews-war/src/main/resources/spring/security/spring-config.xml
```diff
@@ -0,0 +1,92 @@
...
+	<oauth:resource-server id="resourceServerFilter" resource-id="wfim" token-services-ref="tokenServices" />
+	
+	 <bean id="basicAuthenticationFilter" class="ca.bc.gov.webade.oauth2.spring.security.authentication.www.BasicAuthenticationFilter">         
+		<constructor-arg index="0" ref="basicAuthenticationManager"/>
+		<constructor-arg index="1" ref="authenticationEntryPoint"/>
+	</bean>
+
+	<bean id="basicAuthenticationManager"
+		class="ca.bc.gov.webade.oauth2.spring.security.authentication.BasicAuthenticationManager">
+		<property name="resourceId" value="wfrm" />
```
dhlevi, line 52: This isn't WFRM

client/wfnews-war/src/main/resources/spring/security/spring-config.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dhlevi, line 1: Ideally, we would get rid of the XML configuration and use class based configuration with Spring 5

client/wfnews-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -8,29 +8,61 @@
 		<param-name>contextConfigLocation</param-name>
 		<param-value>
 			classpath:spring/spring-config.xml
+			classpath:spring/security/spring-config.xml
 		</param-value>
 	</context-param>
 
+	<filter>
+		<filter-name>springSecurityFilterChain</filter-name>
```
dhlevi, line 16: Ensure this doesn't affect public access.

dhlevi (changes requested):
Make sure the configurations are correct (WFRM is incorrect). I would prefer using beans in classes vs XML configs, but we can update that later if things are working with the old style configs for now.

PR #191, opened 2022-09-22
use environment viriable instead of hardcoded in properties file

yzlucas:
I set: 
oauth2 stuffs ,  
WFIM_API_URL, 
ORG_UNIT_URL, 
WFDM_API_URL ,
WFNEWS_API_URL

anything else should be environment variables?


dhlevi (changes requested):
Make sure to clean out the Application.properties and static.properties of variables we don't use there anymore, or we may get unexpected issues reading properties

yzlucas:
>

PR #200, opened 2022-09-23
Wfnews 494 and snack bar duration

yzlucas:
stucked with servlet mapping issue.... so checking in some WFNEWS-494 and WFNEWS-507 tickets


yzlucas:
SonarCloud complaines about duplicate lines of code, But their snackBar message are different, do we still consider them as duplicate?

dhlevi (approved):

PR #211, opened 2022-09-26
Feature/wfnews 375

ssylver93:


dhlevi (approved):
We'll have to review the warnings and update soon

PR #215, opened 2022-09-26
Fix/wfnews 508

gpineiro-vivid:


client/wfnews-war/src/main/angular/src/app/app.routing.ts
```diff
@@ -35,4 +35,4 @@ const PANEL_ROUTES: Routes = [
 ];
 
 
-export const ROUTING = RouterModule.forRoot(PANEL_ROUTES, { relativeLinkResolution: 'legacy' });
+export const ROUTING = RouterModule.forRoot(PANEL_ROUTES, { relativeLinkResolution: 'legacy', useHash:true });
```
dhlevi, line 38: We shouldn't need the `useHash` here. We have routing handled in Tomcat so the workaround isn't needed on our dev deploy

client/wfnews-war/src/main/angular/package.json
```diff
@@ -14,8 +14,8 @@
     "lint": "ng lint",
     "e2e": "ng e2e",
     "cr": "npm run lint && npm start.prod",
-    "mavenbuild": "node --max_old_space_size=4096 node_modules/@angular/cli/bin/ng build --configuration production --output-path ../../../target/angular --base-href / && replace-in-file /ngsw:/g ngsw:wfnewsi: dist/ngsw-worker.js --verbose --isRegex",
-    "mavenbuild-source-map": "node --max_old_space_size=4096 node_modules/@angular/cli/bin/ng build --configuration production --output-path ../../../target/angular --base-href / --source-map && replace-in-file /ngsw:/g ngsw:wfnewsi: ../../../target/angular/ngsw-worker.js --verbose --isRegex"
+    "mavenbuild": "node --max_old_space_size=4096 node_modules/@angular/cli/bin/ng build --configuration production --output-path ../../../target/angular --base-href ./ && replace-in-file /ngsw:/g ngsw:wfnewsi: dist/ngsw-worker.js --verbose --isRegex",
```
dhlevi, line 17: Dev deploy should stick with the / I beleive

dhlevi (changes requested):
Couple of small conflicts and it looks like a checked in part on the routing that we won't need for dev builds, otherwise appears to be in place.

I would like to update from using xml configs for spring to class and annotation based configs in the future, but we need to be unstuck on testing this now.

PR #263, opened 2022-10-01
WFNEWS-375 Update spring security for WFNEWS top level scope to avoid 403 error

ssylver93:


dhlevi (approved):
What is the value in Scopes.GET_TOPLEVEL?

I think we'd want to make sure it's WFNEWS.GET_TOPLEVEL

PR #300, opened 2022-10-12
Wfnews 258 calculate distance

yzlucas:


dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/panel-wildfire-stage-of-control/panel-wildfire-stage-of-control.component.ts
```diff
@@ -79,4 +93,30 @@ export class PanelWildfireStageOfControlComponent extends CollectionComponent im
           }
     }
 
+    calculateDistance(lat: string, long: string){
```
dhlevi, line 96: Best to pass in Numbers, rather than casting internally. A calculate function would expect numeric latitudes and longitudes

client/wfnews-war/src/main/angular/src/app/components/panel-wildfire-stage-of-control/panel-wildfire-stage-of-control.component.ts
```diff
@@ -79,4 +93,30 @@ export class PanelWildfireStageOfControlComponent extends CollectionComponent im
...
+        if ((Number(lat) && Number(long)) && (this.currentLat && this.currentLong)) {
+            let result = (this.distanceHelper(Number(lat), Number(this.currentLat), Number(long), Number(this.currentLong)))
+            return result.toString() + ' KM'
+        }
+    }
+    
+    distanceHelper(lat1, lat2, lon1, lon2) {
```
dhlevi, line 103: Glad to see the GCA!

It would be better to stick with metres (far more accurate) and convert to KM's / round from there.

```
const R = 6371e3; // metres
const φ1 = lat1 * Math.PI / 180; // φ, λ in radians
const φ2 = lat2 * Math.PI / 180;
const Δφ = (lat2 - lat1) * Math.PI/180;
const Δλ = (lon2 - lon1) * Math.PI/180;

const a = Math.sin(Δφ/2) * Math.sin(Δφ/2) +
          Math.cos(φ1)   * Math.cos(φ2)   *
          Math.sin(Δλ/2) * Math.sin(Δλ/2);
const c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));

const d = R * c; // in metres
```

Above is the haversine formula. If you want a 1-liner that's also pretty accurate, you can look up the spherical law of cosines

dhlevi: Also, this is the kind of thing that should be moved into a Util class, in case we need it elsewhere

PR #324, opened 2022-10-18
Wfnews 373

yzlucas:
change required scope on admin to WFNEWS_ADMIN


client/wfnews-war/src/main/angular/src/app/shared/scopes/scopes.ts
```diff
@@ -1,4 +1,4 @@
-export const SCOPE_APPLICATION_PREFIX = 'WFIM.';
+export const SCOPE_APPLICATION_PREFIX = 'WFNEWS.';
```
dhlevi, line 1: We might have mixed WFIM and WFNEWS scopes used. Worth double checking we're using the right ones here

dhlevi (approved):

PR #337, opened 2022-10-21
add searchText for publicIncidents

yzlucas:


server/wfnews-persistence/src/main/resources/ca/bc/gov/nrs/wfnews/persistence/v1/dao/mybatis/mapper/PublishedIncidentMapper.xml
```diff
@@ -248,6 +248,16 @@
 			<if test="fireCentre != null &amp;&amp; fireCentre != ''">  
 				AND fire_centre = #{fireCentre, javaType=java.lang.String, jdbcType=VARCHAR, mode=IN}
 			</if>
+			<if test="searchText!=null &amp;&amp; searchText.length>0">
+				AND (
+					<foreach item="item" index="index" collection="searchText" separator="AND" >
+						( UPPER( 	 incident_name||'~^~'|| incident_number_label
```
dhlevi, line 254: This query will never return a value.

You do not want to use `||` to concatenate possibly null values, or it will always return null. Use `CONCAT(<val>, '<delimeter>', <val2>)` or `Coalesce` (though concat will do the trick for us in postgres, it handles null values)

Change to:

```
AND UPPER(CONCAT(incident_location, ' ', CONCAT(fire_centre, ' ', CONCAT(incident_number_label, ' ', incident_name))))
LIKE '%'||UPPER('<my value>')||'%'
```

Alternatively, you could try using text search
Replace this with:

`AND to_tsvector(CONCAT(incident_location, ' ', CONCAT(fire_centre, ' ', CONCAT(incident_number_label, ' ', incident_name))))
    @@ to_tsquery(<my string to search>)`

And pass in a string rather than the array.

You can allow the user to then pass in more robust queries that can include operators such as negate `!` and `&` or `|`

You'd want to test this a bit to see if it meets the needs. Text search is phrase, not wildcard, so if we need partial word LIKE is probably better. Either way, it only takes about 5 minutes to get a local environment set up so you can test this out easy

dhlevi (changes requested):
There is an issue with the query (see comment) but otherwise no issues.

Generally we add new parameters to the "back" of the list, rather than the front.

server/wfnews-persistence/src/main/resources/ca/bc/gov/nrs/wfnews/persistence/v1/dao/mybatis/mapper/PublishedIncidentMapper.xml
yzlucas: Thank you. 
I have updated the query and tested by hitting local api with Postman, Verify the searchText works on those columns

dhlevi (approved):

PR #342, opened 2022-10-25
Feature/wfnews 295

gpineiro-vivid:


dhlevi (approved):

dhlevi:
Looks good. Only issue I see is that the icons dont change the pointer so it isn't obvious where to click for downloading, but that's a quick css fix we can apply later

PR #345, opened 2022-10-25
WFNEWS-544: Update search box styling

dhlevi:


client/wfnews-war/src/main/angular/src/styles/_base.scss
```diff
@@ -596,17 +596,18 @@ iframe#login-iframe {
     border-right: none;
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
+    border-right: none;
```
yzlucas, line 599: sonar bug: duplicate "border-right: none;"

yzlucas (approved):

client/wfnews-war/src/main/angular/src/styles/_base.scss
dhlevi: whoops! Will update

PR #354, opened 2022-10-27
Wfnews 284

yzlucas:
current notices list done.


client/wfnews-war/src/main/angular/src/app/components/current-alert-component/current-alert-component.component.html
```diff
@@ -52,7 +52,66 @@
...
+                <span class="item-title">Fire Centre: </span>
+                <span class="item-content">{{ban.fireCentre}}</span>
+            </div>
+            <div class="item-row">
+                <span class="item-title">Restrictions: </span>
+                <span class="item-content">{{convertRestriction(ban.accessProhibitionDescription)}}</span>
+            </div>
+            <div class="item-row">
+                <span class="item-title">Distance from My Location: </span>
+                <span class="item-content">{{calculateDistance(ban.centroid.y, ban.centroid.x)}} KM</span>
```
dhlevi, line 92: minor niggle, but I think we used lowercase km elsewhere. We can double check for consistency later

client/wfnews-war/src/main/angular/src/app/utils/index.ts
```diff
@@ -2,6 +2,7 @@ import { APP_BOOTSTRAP_LISTENER, Inject, InjectionToken, Type } from '@angular/c
 import { EffectSources } from '@ngrx/effects';
 import { SortDirection } from '@wf1/core-ui';
 import * as moment from 'moment';
+import { Moment } from 'moment';
```
dhlevi, line 5: You have Moment imported directly above this. No need to load it twice.

Side note, preferably we don't use the moment library at all. It really isn't needed

dhlevi (approved):

PR #366, opened 2022-10-28
Non comp

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/panel-wildfire-stage-of-control/panel-wildfire-stage-of-control.component.ts
```diff
@@ -384,4 +384,7 @@ export class PanelWildfireStageOfControlComponent extends CollectionComponent im
       console.log(event)
       event.stopPropagation()
     }
+
+    disableWildfiresOfNoteCheckbox () {
```
dhlevi, line 388: Any reason this is empty?

dhlevi (approved):
If that empty function is not needed, it should be removed

client/wfnews-war/src/main/angular/src/app/components/panel-wildfire-stage-of-control/panel-wildfire-stage-of-control.component.ts
yzlucas: sorry, committed it in by mistake,

PR #405, opened 2022-11-10
Current Notices displayed stacked in single column in desktop mode again.

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/current-alert-component/current-alert-component.component.scss
```diff
@@ -154,7 +154,7 @@ select{
 
 
 
-@media (min-width: 768px) {
+@media (max-width: 820px) {
```
dhlevi, line 157: I feel like this will be a constant source of difficultly... might be better to adjust to rem in case browser pixel sizes have been scaled, but we can work on testing that later

dhlevi (approved):

PR #412, opened 2022-11-15
WFNEWS-598

gpineiro-vivid:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
```diff
@@ -52,14 +52,17 @@ export class IncidentInfoPanel implements AfterViewInit {
 
   public printPage() {
     const printContents = document.getElementsByClassName('page-container')[0].innerHTML
-    const originalContents = document.body.innerHTML
+    
+    var appRoot = document.body.removeChild(document.getElementById("app-root"));
```
dhlevi, line 56: can be `const` rather than var, unless you want it globally scoped

dhlevi (approved):

PR #431, opened 2022-11-21
597 628

yzlucas:


client/wfnews-war/src/main/angular/src/app/utils/index.ts
```diff
@@ -25,42 +25,54 @@ export const FireCentres = [
         description: "Cariboo Fire Centre",
         displayOrder: 1,
         effectiveDate: "1999-01-01",
-        expiryDate: "9999-12-31"
+        expiryDate: "9999-12-31",
+        integerAlias: 7,
```
dhlevi, line 29: Integer Aliases are by Fire Zone. Each centre will have many of them. It might be easier to have a mapping by fire zone org unit in this case

dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/utils/index.ts
yzlucas: we probably dont need integerAlias in our case, can I remove this?

dhlevi: Yeah, if we're not using it may as well remove it

PR #462, opened 2022-11-29
Wfnews 689

yzlucas:


dhlevi (approved):
Looks good, but did you verify that this didnt affect any other place where the mat-label-active class is used? the :host and ::ng-deep can do unexpected things depending on load order

yzlucas:
> Looks good, but did you verify that this didnt affect any other place where the mat-label-active class is used? the :host and ::ng-deep can do unexpected things depending on load order

Yes, and all the active-tabs should apply this styling actually.

PR #466, opened 2022-11-29
Ability to add evac orders as URI's

dhlevi:


yzlucas (approved):
looks good, although a bit hacky to use externalUri to store these evac data.

PR #519, opened 2023-01-09
WFNEWS-751: Add response type, mappings for descriptions

dhlevi:
Adding response type label, descriptions to Wildfire Information section on public incident view below stage of control and cause code.

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
@@ -62,6 +62,12 @@ export class IncidentInfoPanel implements AfterViewInit {
     else return 'A wildfire of undetermined cause, including a wildfire that is currently under investigation, as well as one where the investigation has been completed.'
   }
 
+  public getResponseTypeDescription (code: string) {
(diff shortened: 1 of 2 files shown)


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
```diff
@@ -62,6 +62,12 @@ export class IncidentInfoPanel implements AfterViewInit {
     else return 'A wildfire of undetermined cause, including a wildfire that is currently under investigation, as well as one where the investigation has been completed.'
   }
 
+  public getResponseTypeDescription (code: string) {
+    if (code === 'MONITOR') return 'When a fire is being monitored, this means BC Wildfire Service is observing and analyzing the fire but it\'s not immediately suppressed. It may be allowed to burn to achieve ecological or resource management objectives and is used on remote fires that do not threaten values.'
```
ssylver93, line 66: Just a typo here for "it's" @dhlevi

yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
dhlevi: Not a Typo! the slash is needed to break the single-quote. I could change the string to double-quotes but that would be inconsistent.

PR #521, opened 2023-01-09
WFNEWS-765: Remove splash screen

dhlevi:
Ticket specified removing title and updating logo, however a secondary request from Briana was "just remove the splash screen". So this is the removal, we can test for startup performance.


yzlucas:
There is `window['SPLASH_SCREEN'].remove();` in app.component.ts 
Shall we also remove that?

dhlevi:
Yeah, I was going to at first, but it's basically a no-op and if we need the splash screen back it's one less thing to forget. Rather than remove, it'd be a good case of commenting out with a note indicating what it's commented out for.

yzlucas (approved):

PR #524, opened 2023-01-09
849: separate searchbox for address

yzlucas:


dhlevi (approved):
Code changes look OK, might need some testing to make sure it's all placed well and behaves as expected

PR #525, opened 2023-01-09
WFNEWS-837: Disable map save with no title, WFNEWS-820

dhlevi:
Disabled save button on map dialog, also small fix to the back to dashboard link for WFNEWS-820

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/maps-panel/upload-map-dialog/upload-map-dialog.component.html
+++ b/client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/maps-panel/upload-map-dialog/upload-map-dialog.component.html
@@ -9,7 +9,7 @@
     </mat-form-field>
   </div>
   <div class="dialog-actions" mat-dialog-actions>
-  	<button mat-raised-button color="primary" [mat-dialog-close]="returnResult()" (disabled)="file === null">Save</button>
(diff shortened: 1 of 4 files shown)


client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/maps-panel/upload-map-dialog/upload-map-dialog.component.html
```diff
@@ -9,7 +9,7 @@ <h1 mat-dialog-title>Upload Map</h1>
     </mat-form-field>
   </div>
   <div class="dialog-actions" mat-dialog-actions>
-  	<button mat-raised-button color="primary" [mat-dialog-close]="returnResult()" (disabled)="file === null">Save</button>
+  	<button mat-raised-button color="primary" [mat-dialog-close]="returnResult()" [disabled]="file === null || title === null || title === ''">Save</button>
```
yzlucas, line 12: probably we can simplify this with "[disabled]="file === null || !title" ?

yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/maps-panel/upload-map-dialog/upload-map-dialog.component.html
dhlevi: Yep that'd be fine for this case, I don't think we're going to change it in the future to maintain empty but detect nulls/undefined

PR #529, opened 2023-01-10
feature/WFNEWS-476 update getters and setters for fileId and thumbId …

ssylver93:
…which are causing null values


dhlevi (approved):
COde changes are good. Are the dependency errors something we can address?

PR #539, opened 2023-01-11
Wfnews-858: Update the incident admin form based on Taylor's incident details functions document

yzlucas:


ssylver93 (approved):

dhlevi (approved):
It might not be accurate enough to just search for the word "estimated" but we can test it out and verify

PR #551, opened 2023-01-17
Wfnews 814

yzlucas:
Add fireCentreName attribute and update api to fetch and orderBy it. I have reviewed with Sean to verify the validation

### DIFF
--- a/database/scripts/01_00_00/05/ddl/tables/MODIFY.PUBLISHED_INCIDENT_DETAIL.sql
+++ b/database/scripts/01_00_00/05/ddl/tables/MODIFY.PUBLISHED_INCIDENT_DETAIL.sql
@@ -1,0 +1,7 @@
+-- Add FIre Centre Name
+ALTER TABLE "wfnews"."published_incident_detail" ADD COLUMN IF NOT EXISTS "fire_centre_name" VARCHAR(120);
+COMMENT ON COLUMN "wfnews"."published_incident_detail"."fire_centre_name" IS 'fire_centre_name is the name of fire centre of the incident'
+;
+
+CREATE INDEX "pubincdtl_fire_centre_name_idx" ON "wfnews"."published_incident_detail" USING btree("fire_centre_name")
+;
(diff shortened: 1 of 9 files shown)


server/wfnews-persistence/src/main/java/ca/bc/gov/nrs/wfnews/persistence/v1/dto/PublishedIncidentDto.java
```diff
@@ -108,6 +109,7 @@ public PublishedIncidentDto(PublishedIncidentDto dto) {
 		this.longitude = dto.longitude;
 		this.declaredOutDate = dto.declaredOutDate;
 		this.fireCentre = dto.fireCentre;
+		this.fireCentreName = dto.fireCentreName;
```
dhlevi, line 112: You added the attribute here, but not in the resource constructor. Add it to the constructor that passes in a PublishedIncident object (line 118)

dhlevi: You also didn't add a check in the `equalsAll` method, which is required for equality checks.

dhlevi (changes requested):
You're missing mappings in PublishedIncidentResourceFactory, which would prevent the new attribute from ever being mapped? Make sure you test API changes, there are Docker scripts set up to assist with this if you don't have a local debug environment for java APIs.

database/scripts/01_00_00/05/ddl/tables/MODIFY.PUBLISHED_INCIDENT_DETAIL.sql
```diff
@@ -0,0 +1,7 @@
+-- Add FIre Centre Name
+ALTER TABLE "wfnews"."published_incident_detail" ADD COLUMN IF NOT EXISTS "fire_centre_name" VARCHAR(120);
+COMMENT ON COLUMN "wfnews"."published_incident_detail"."fire_centre_name" IS 'fire_centre_name is the name of fire centre of the incident'
```
dhlevi, line 3: The comment on the existing column `"wfnews"."published_incident_detail.fire_centre"` should be updated to reflect that it's only storing the Code. Additionally, now that we have a fire_centre_name, column, that old column should be renamed to fire_centre_code or something sensible to reflect what it now is.

yzlucas:
> You're missing mappings in PublishedIncidentResourceFactory, which would prevent the new attribute from ever being mapped? Make sure you test API changes, there are Docker scripts set up to assist with this if you don't have a local debug environment for java APIs.

PR #555, opened 2023-01-18
WFNEWS-739

gpineiro-vivid:


dhlevi:
The conflict in the code appears to be attempting to change the code in a nearly identical way as the current code... Can you make sure to pull the latest and ensure the change you have here is still valid?

dhlevi (changes requested):
See attached comment

gpineiro-vivid:
@dhlevi the change seems to be no longer valid, the issue was solved and seems like I did not have the latest change, sorry. Moving the issue to Ready for testing to double check,

PR #556, opened 2023-01-18
Non compliance

dhlevi:
Update for weather layer to ensure it will not display if turned off, hide the legend.

Update to published incident GET to replace out boolean with list of stages of control
Update to the UI to pass in a list of stages of control

Update to hide youtube video box on print preview

Adding character limits and display to admin page video, image and pdf upload/edit


dhlevi:
This has been sitting here for more than 24 hours without a review. I'll merge it in and hopefully nothing explodes!

PR #558, opened 2023-01-19
add disclaimer pop out

yzlucas:


dhlevi (approved):

dhlevi:
The security warnings are related to noopeners on the gov links. We don't need them in that case so we can let it slide. Feel free to add them if you want but I dont think it's required in this case

PR #561, opened 2023-01-20
Wfnews 862, add default content for current resources

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.html
```diff
@@ -99,91 +99,96 @@ <h1>Response</h1>
...
+    <div *ngIf="(incident.wildfireCrewResourcesInd || incident.wildfireAviationResourceInd || incident.incidentMgmtCrewRsrcInd || incident.heavyEquipmentResourcesInd || incident.structureProtectionRsrcInd); else defaultResourcesContent">
+      <!-- Wildfire Resources -->
+      <div *ngIf="incident.wildfireCrewResourcesInd" class="resource-panel">
+        <div class="resource-panel-icon">
+          <img class="image-size" src="../../../assets/images/svg-icons/wildfire-crews.svg" alt="Wildfire Crews" />
+        </div>
+        <div class="resource-panel-content">
+          <h3>Wildfire Crews</h3>
+          <div class="content-panel-body">
+            <div [innerHTML]="incident.wildfireCrewResourcesDetail || 'There are currently Initial Attack and Unit Crews responding to this wildfire.'"></div>
```
dhlevi, line 116: So, This message will show only if the Indicator is set to true. If the indicator is true, the default message doesnt make any sense. this should only display if the indicator is FALSE, not if there is no message.

This comment is consistent with the other sections here as well. It should only be showing this message if the IND is FALSE, not if the IND is TRUE but there is no set details.

dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.html
yzlucas: Right, 
The current rule is only shows the resposne detail content when the IND is ture, and there is set details" . 
we will not show "There are currenlty xxx...." content without number in it.  

Please correct me if I miss something here...

dhlevi (approved):

PR #562, opened 2023-01-20
Non compliance remediation srpint, avoid auto focusing of first element in dialog, and wfnews-873

yzlucas:


dhlevi (approved):
Changes look good.

Be careful when adding Divs to wrap a section for an ngIf, in case the Divs have default styling applied by a parent container. It could cause the layout to mess up unexpectedly

PR #563, opened 2023-01-23
Non compliance remediation sprint2

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.ts
```diff
@@ -104,4 +104,16 @@ export class IncidentHeaderPanel implements AfterViewInit {
   removeFromWatchlist () {
     this.watchlistService.removeFromWatchlist(this.incident.incidentNumberLabel)
   }
+
+  displaySizeType (incidentSizeDetail: string) {
+    if (incidentSizeDetail && incidentSizeDetail.includes('estimated')) {
+      return '(Estimated)'
```
dhlevi, line 110: Would probably make sense to move the brackets out into the HTML rather than part of the output here

client/wfnews-war/src/main/angular/src/app/components/incident-identify-panel/incident-identify-panel.component.scss
```diff
@@ -53,6 +53,9 @@
     .label {
       color: #036;
     }
+    .paragraph-format{
```
dhlevi, line 56: We already have a paragraph-format class with the same settings. You might be able to get this from a parent scss without the addition here

dhlevi (approved):

PR #566, opened 2023-01-25
WFNEWS-874

gpineiro-vivid:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
```diff
@@ -16,19 +17,26 @@ export class IncidentInfoPanel implements AfterViewInit {
...
   public primaryMedia = null
 
+
   ngAfterViewInit(): void {
     const canvas = document.getElementById('qr-code')
     toCanvas(canvas, window.location.href, function (error) {
       if (error) console.error(error)
     })
 
+    this.showWarning = this.evacOrders.length > 0 || this.areaRestrictions.length > 0;
```
dhlevi, line 38: The boolean for showWarning shoul only display if params['preview'] is true? Not sure what this logic is doing?

Most of this could be replaced with just an ngIf params['preview']?

gpineiro-vivid: It was my understanding that we need to show the warnings in "preview" mode IF the incident got images, maps, evacuation orders, etc. Is this correct? Anyways, I need to change the ngIf validating both "showWarning" and "preview"

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
dhlevi: True, however I'm not seeing how the logic here takes into account preview mode. That's my only concern here is that the label will display in the regular view when the collection is empty, which is not what we want, hence have an explicit check on preview mode for safety.

gpineiro-vivid: @dhlevi done

PR #579, opened 2023-01-25
WFNEWS-670: Add 160px of spacing

dhlevi:
This amount of spacing, regardless of meeting the TAC, looks terrible. Strongly recommending we don't use this


yzlucas (changes requested):
padding is too big... But I see Jacqueline mentioned in the ticket to change this to 40px for each spacer.  Could be better

dhlevi:
Yeah, I put in the full 80 per the TAC, but it's really big. After discussion with Jacky we're thinking of cutting it down a bit

PR #602, opened 2023-02-01
Wfnews 860

yzlucas:
also trigger searchTextUpdated after deleting keyword from location searchbox


dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -271,4 +271,23 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
...
+  }
+
+  isSearchText() {
+    if (this.searchText && this.searchText !== ""){
+      return true
+    }
+  }
+
+  locationNameUpdated() {
+    if (this.locationName == '') {
```
dhlevi, line 287: ===

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -271,4 +271,23 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
...
   }
 
+  isLocationName() {
+    if (this.locationName && this.locationName !== ""){
+      return true
+    }
+  }
+
+  isSearchText() {
+    if (this.searchText && this.searchText !== ""){
```
dhlevi, line 281: consistent with the quotes. Use single or double, but try not to mix both

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -271,4 +271,23 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
     this.doSearch()
   }
 
+  isLocationName() {
+    if (this.locationName && this.locationName !== ""){
```
dhlevi, line 275: just return the conditional:

`return this.locationName && this.locationName !== ''`

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -271,4 +271,23 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
...
+
+  isSearchText() {
+    if (this.searchText && this.searchText !== ""){
+      return true
+    }
+  }
+
+  locationNameUpdated() {
+    if (this.locationName == '') {
+      this.selectedLat=undefined;
```
dhlevi, line 288: I don't mind if you don't use semi-colons (not required in typescript outside of specific cases) but don't mix them up or it starts to look weird. Also, it's nice to format things so there's spaces between symbols

`this.selectedLong = undefined` is more readable

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
yzlucas: return this.locationName && this.locationName !== ''  doesn't really work here.
when this.locationName = '', return ( this.locationName && this.locationName !== '' ) will not return a boolean. 
It is actually return ( '' & false) , and the result is not false

dhlevi: Right. You'd have to define the conditional to get rid of the behaviour when using `if (this.locationName)`

PR #607, opened 2023-02-02
Disclaimer link button mobile

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -492,4 +489,12 @@ export class ActiveWildfireMapComponent implements OnInit, AfterViewInit {
       window.open('https://www2.gov.bc.ca/gov/content/home/copyright', "_blank");
     }
   }
+
+  disclaimerText() {
+    if (screen.width <= 1200) {
+      return 'Links';
```
dhlevi, line 495: Check with Paul on wording here if you haven't already. "Disclaimer" might make more sense, but obviously space gets tight unless we move over the other header boxes.

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
yzlucas: Paul suggests a modification to "Legal"

dhlevi: Sounds good

PR #625, opened 2023-02-07
add a radio button to unset a primary video

yzlucas:


dhlevi (approved):
If the API handles un-setting the primary, this shouldn't be needed, but we can implement for now. We may roll it back if the API change makes it in on WFIM

PR #640, opened 2023-02-16
add more fields to /feature

yzlucas:


server/wfnews-persistence/src/main/resources/ca/bc/gov/nrs/wfnews/persistence/v1/dao/mybatis/mapper/PublishedIncidentMapper.xml
```diff
@@ -417,7 +417,7 @@
             'geometry',   ST_AsGeoJSON(t.geometry, options => 1)::JSONB,
             'properties', TO_JSONB(t.*) - 'geometry'
            ) AS feature
-		          FROM (SELECT f.incident_number_label, f.incident_name, f.fire_year, f.geometry FROM wfnews.published_incident_detail AS f
+		          FROM (SELECT f.incident_number_label, f.incident_name,f.incident_location, f.discovery_date, f.general_incident_cause_cat_id, f.incident_size_estimated_ha, f.stage_of_control_code, f.incident_mgmnt_crew_rsrc_ind, f.wildfire_aviation_resource_ind, f.wildfire_crew_resources_ind, f.heavy_equipment_resources_ind, f.structure_protection_rsrc_ind, f.last_updated_timestamp, f.fire_year, f.geometry FROM wfnews.published_incident_detail AS f
```
dhlevi, line 420: So, I'm getting a little nervous about this much data coming back in the spatial call. It is a big payload already and we deliberately wanted to limit this to ID's.

Is there no way for PM to get the ID and call the API directly for the incident?

yzlucas: Because we are showing this incident detail in the map layer. (PM tile) I am not sure how do I call the API directly and still reflect the info in layers. What I am trying to do is keep as consitency as possible to the data that come back from AGOL server.  
This is quite different as our NEWS' pop out incident detail panel.

dhlevi: It should be more or less the same? Take a look at the weather panel and how it calls the pointid API and loads data there.

PR #694, opened 2023-04-27
change to use NEWS_CREATED_TIMESTAMP to make sure not fetching update…

yzlucas:
…d incident in monitor-api


dhlevi (changes requested):
Firstly, you have adjusted "from", but not "to" dates. We would need to change both for a valid test
Secondly, this check appears in two locations, so you will need to adjust both areas the check is done (once on select, once on count)

Final thought... The param createDate expects to query "createDate" not "newsCreatedTimestamp"... we might want to consider renaming the param to be consistent with what we're querying, but because both should be close to the same (WFIM Create date and WFNEWS create date should realistically be almost exactly the same, maybe off by a few seconds) it probably doesn't really matter.

yzlucas:
> createDate

Those two timestamp usually are off by few millisecond. I didn't rename the param because I think we still need to find the issue in syncer and may change back to use Create_Time eventually.

dhlevi:
Change looks ok, waiting to hear if there'll be a change on the syncer before commiting

PR #695, opened 2023-04-28
Adding missed dates in the dto layer causing new createDate to be set…

ssylver93:
… on update


dhlevi (commented):
I think there are additional places to update?
Verify that we don't need to add checks in the Factory class

ssylver93:
> I think there are additional places to update? Verify that we don't need to add checks in the Factory class

@dhlevi confirmed that these changes are the only ones needed in the news API. The createDate and updateDate are already getting set correctly in the factory class when I test the change locally.

PR #699, opened 2023-05-03
Feature/wfnews 913 - Return matching address at top of address list

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -91,22 +91,42 @@ export class ActiveWildfireMapComponent implements OnInit, AfterViewInit {
...
               result.address = address.trim();
               self.highlight(result);
-            });
+              trimmedAddress = result.address;
+              valueLength = val.length;
+              if (trimmedAddress != null) valueMatch = trimmedAddress.substring(0, valueLength);
+
+              if (address != null && valueLength != null && valueMatch != null &&
+                (val.toUpperCase() === address.toUpperCase() || val.toUpperCase() === valueMatch.toUpperCase())) {
+                  var index = results.indexOf(result);
```
dhlevi, line 111: replace the var with a let or const (I assume const in this case as that value isn't changing)

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -81,20 +81,38 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
...
+            results.forEach((result) => {
+              address = self.getFullAddress(result);
+              result.address = address.trim();
+              trimmedAddress = result.address;
+              valueLength = val.length;
+              if (trimmedAddress != null) valueMatch = trimmedAddress.substring(0, valueLength);
+
+              if (address != null && valueLength != null && valueMatch != null &&
+                (val.toUpperCase() === address.toUpperCase() || val.toUpperCase() === valueMatch.toUpperCase())) {
+                  var index = results.indexOf(result);
```
dhlevi, line 102: Replace the var with let or const (as above, I assume const here as it doesnt change)

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -479,11 +499,11 @@ export class ActiveWildfireMapComponent implements OnInit, AfterViewInit {
     this.resizeBoxElement.style.height = `${window.innerHeight - this.lastPointerPosition + 20}px`
   }
 
-  openLink(link:string) {
+  openLink(link: string) {
     if (link === 'Disclaimer') {
       window.open('https://www2.gov.bc.ca/gov/content/home/disclaimer', "_blank");
```
dhlevi, line 504: Not related to this PR, but those URLs should likely be added to the configuration rather than being hardcoded

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -479,11 +499,11 @@ export class ActiveWildfireMapComponent implements OnInit, AfterViewInit {
...
     }
-    else if  (link === 'Privacy') {
+    else if (link === 'Privacy') {
       window.open('https://www2.gov.bc.ca/gov/content/home/privacy', "_blank");
```
dhlevi, line 507: Not related to this PR, but those URLs should likely be added to the configuration rather than being hardcoded

client/wfnews-war/src/main/angular/src/app/components/wildfires-list/wildfires-list.component.ts
```diff
@@ -81,20 +81,38 @@ export class WildFiresListComponent extends CollectionComponent implements OnCha
...
-          this.placeData.searchAddresses(val).then(function(results){
-              if(results) {
-
-                  results.forEach((result) => {
-                      let address = self.getFullAddress(result);
-                      result.address = address.trim();
-                  });
-
-                  self.filteredOptions = results;
+      if (val.length > 2) {
```
dhlevi, line 84: A lot of duplicate code here.
Can the code block from here and from the `active-wildfire-map` component move the duplicates into a service or utility function so we only need the code in one place?

dhlevi (changes requested):

dhlevi (approved):

PR #701, opened 2023-05-08
wfnews-915

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-panel/incident-info-panel.component.ts
```diff
@@ -57,7 +57,7 @@ export class IncidentInfoPanel implements AfterViewInit {
 
     this.router.queryParams.subscribe((params: ParamMap) => {
       if(params['preview'])
```
dhlevi, line 59: params['preview'] is returning a boolean I beleive, so you can probably just do `this.showWarning = params['preview']`

dhlevi (approved):

PR #714, opened 2023-05-12
Install Ionic & Capacitor 5, add plugins, capacitor service, update geolcation for mobile

ssylver93:


client/wfnews-war/src/main/angular/android/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,71 @@
...
+    <uses-feature
+        android:name="android.hardware.camera"
+        android:required="false" />
+    <!-- Network API -->
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <!-- Navigator.getUserMedia -->
+    <!-- Video -->
+    <uses-permission android:name="android.permission.CAMERA" />
+    <!-- Audio -->
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
```
dhlevi, line 68: I dont think we need this permission?

dhlevi (changes requested):
Lots of sonar failures here to address

PR #715, opened 2023-05-12
Size and logging fix

vivid-cpreston:


.github/workflows/deploy-prod.yml
cboreen, line 1: Did somebody decide the new max?

dhlevi: Not that I know of. 20mb seems pretty large. 5-10 is probably safer? (default is 10 I see in your config below)

dhlevi (approved):
We may want to adjust the size in the future. 20 seems large today, but when we add video, etc, it'll be too small. WFDM controls the limit technically so we might want to consider leaving it open

PR #725, opened 2023-05-31
wfnews-1137

yzlucas:


dhlevi (approved):
Nothing wrong with the changes here, but a thought:

Now that we're going to officially be supporting mobile devices, would it make more sense to use explicit mobile html pages rather than size detection on the single page?

Might be worth a team discussion. Personally I dislike having things too separate because it leads to a load of code duplication, but as long as we're keeping the ts and css together it should be ok

yzlucas:
> Nothing wrong with the changes here, but a thought:
> 
> Now that we're going to officially be supporting mobile devices, would it make more sense to use explicit mobile html pages rather than size detection on the single page?
> 
> Might be worth a team discussion. Personally I dislike having things too separate because it leads to a load of code duplication, but as long as we're keeping the ts and css together it should be ok



I am working on seperate them... ie: incident-info-panel-mobile component which will extends IncidentInfoPanelMobileComponent, basically an empty ts, but unique HTML and CSS files.  I am not finish yet but I will create a PR of what I have on it , And let me know what do you think of my approach

dhlevi:
OK sounds good. I'll likely wait to merge both together then so we don't break the screen in case other testing is going on

dhlevi:
@yzlucas Looks like the recent merge has resulted in some conflicts here. Verify and cleanup and we can merge this in as well

PR #730, opened 2023-06-01
wip of overview tab (separate as mobile component) 

yzlucas:


yzlucas:
Still in progress, but just want to show what I am doing here.... Basically creating mobile component for each of 4 tabs. I think it would be easier to manage the html and css code

ssylver93 (approved):

PR #735, opened 2023-06-09
Remove byte array and content length header to use ResponseInputStrea…

ssylver93:
…m for attachment GET


dhlevi:
Are there any addition changes for this or good to push to main as well?

dhlevi:
I think it might need the bugfix branch merged in first? @ssylver93

dhlevi:
Closing. Merged with a separate PR

PR #773, opened 2023-06-20
Gallery tab changes

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-gallery-videos-mobile/incident-gallery-videos-mobile.component.scss
```diff
@@ -0,0 +1,44 @@
+.title-row {
+    width: 182px;
+    height: 24px;
+    background-color: rgba(255, 255, 255, 0);
+    box-sizing: border-box;
+    font-family: "HelveticaNeue-Medium", "Helvetica Neue Medium", "Helvetica Neue", sans-serif;
```
dhlevi, line 6: We must use BCSans fonts only

dhlevi (changes requested):
Couple things:
1.) Can we move all of the "gallery" stuff into a subfolder so it's grouped together and doesn't clutter up under the `public-incident-page` folder
2.) There's a lot of duplication going on between the "all" and separate video and image panels. We should find a way to share the code so we're not updating identical stuff between the desktop and mobile variants in multiple places

dhlevi:
@ssylver93 Looking better. There are a few conflicts on the branch that'll need to get resolved first

dhlevi (approved):

PR #778, opened 2023-06-22
WFNEWS-993 : Mobile build automation

sukhpalp:
- Created GitHub Action Workflows
- Changed some Angular configuration to allow for dev builds
- Removed Capacitor plugin that is not required now that the latest is released
- Removed the platform folders, updated the package names and reran the `capacitor add` command
- Ran the projects on the build mac and adjusted settings in their native IDEs to update the platform folders


dhlevi (approved):
Looks good, but there's a lot of changes here. Wouldn't mind @vivid-cpreston taking a look at some of the workflow changes at least for a verification

dhlevi:
@vivid-cpreston Can you please review this PR?

.github/workflows/ios.yml
```diff
@@ -0,0 +1,163 @@
...
+        description: 'Upload to App Store'
+        type: boolean
+        required: true
+  workflow_dispatch:
+    inputs:
+      build_environment:
+        description: 'Build Environment'
+        type: choice
+        options:
+          - development
```
vivid-cpreston, line 21: Are all the non-prod environments being treated as equivalent?

vivid-cpreston (commented):
Although I'm not familiar with everything being run here, it looks good from what I can tell - only question I have is to double-check how it handles differing build environments,

PR #786, opened 2023-06-27
WFNEWS-1197 : Adjust styling to match mockups

sukhpalp:


dhlevi (approved):
Some of the padding and margin changes are pretty big, might have to verify we don't inadvertently adjust the desktop mode in a few spots as well

dhlevi:
@sukhpalp  Looks like there's a merge conflict with a previous change to look at, otherwise changes look OK, but I will want to verify desktop is unaffected by some of the margin adjustments

PR #791, opened 2023-06-29
WFNEWS-1158 Use BufferedReader to convert s3Object bytesResponse and avoid 500 error on image download

ssylver93:


dhlevi:
If we're going to put the response into a string, it would be better to use the byte array

dhlevi (approved):

PR #800, opened 2023-07-04
WFNEWS-1089

dhlevi:
Adding Lambda code, Lambda Layer from WFONE
Adding terraform scripts for setting up Lambda, SQS, S3


dhlevi:
Still some TODO here.
- Add environment vars
- Configure secrets stuff (if we still need it?)
- Role configuration

terraform/sqs.tf
```diff
@@ -0,0 +1,361 @@
+resource "aws_sqs_queue" "deadletter" {
```
vivid-cpreston, line 1: This resource can be removed, is no longer needed in current version

terraform/sqs.tf
```diff
@@ -0,0 +1,361 @@
...
+  name = "wfnews-notification-api-deadletter-queue-${var.target_env}"
+
+  tags = {
+    Application = "wfnews"
+    Customer    = "BCWS"
+    Environment = var.target_env
+  }
+}
+
+resource "aws_sqs_queue" "queue" {
```
vivid-cpreston, line 11: This resource can be removed, is no longer needed in current version

terraform/variables.tf
```diff
@@ -569,4 +569,80 @@ variable "agolBansAndProhibitions" {
...
+  type    = string
+  default = ""
+}
+
+variable "EvacuationMonitorKey" {
+  type    = string
+  default = ""
+}
+
+variable "TwitterMonitorKey" {
```
vivid-cpreston, line 640: No longer needed in current version

vivid-cpreston (commented):
Terraform components look good, just a couple elements that can be cleaned-up/removed as they are no longer needed in current version

dhlevi:
Thanks @vivid-cpreston I'll remove the deprecated components and resubmit

PR #809, opened 2023-07-06
WFNEWS-1081 notification-monitors-api merge

ssylver93:


dhlevi:
Duplicated by https://github.com/bcgov/nr-bcws-wfnews/pull/800

The other PR includes the monitor lambdas as well as the updated terraform scripting, so I think we can keep that one

PR #813, opened 2023-07-06
WFNEWS-1079 wfone-notifications-push-api merge

ssylver93:


server/wfone-notification-push-api/wfone-notification-push-api-rest-endpoints/src/main/resources/firebase-config.json
```diff
@@ -0,0 +1,12 @@
+{
```
dhlevi, line 1: We can remove the client ID and URL info and move that to configurations/secrets, that would be ideal. Rather not have these checked in

dhlevi (changes requested):

server/wfone-notification-push-api/wfone-notification-push-api-rest-endpoints/src/main/resources/firebase-config.json
ssylver93: Thanks I've done so now

PR #852, opened 2023-07-13
notifications-db terraform

yzlucas:


terraform/ecs.tf
vivid-cpreston, line 1: Also needs a service resource to tell it to actually run the task

.github/workflows/terragrunt-deploy.yml
```diff
@@ -254,6 +254,7 @@ jobs:
           GITHUB_RELEASE_NAME: ${{ github.event.release.name }}
           INSTANCE_COUNT: ${{vars.INSTANCE_COUNT}}
           LIQUIBASE_IMAGE: ${{vars.REGISTRY}}/${{ github.repository_owner }}/${{ vars.LIQUIBASE_IMAGE }}:${{ inputs.IMAGE_TAG }}
+          NOTIFICATIONS_LIQUIBASE_IMAGE: ${{inputs.registry}}/${{ github.repository_owner }}/${{ var.NOTIFICATIONS_LIQUIBASE_IMAGE }}:${{ inputs.IMAGE_TAG }}
```
vivid-cpreston, line 257: I don't believe the image you're using here is being created yet - is someone else adding it in a different PR, or do we need to add the github action to do so?

vivid-cpreston (changes requested):

PR #854, opened 2023-07-14
Wfnews 1074 update notifications-db terraform

yzlucas:


yzlucas:
we may need "NOTIFICATIONS_LIQUIBASE_IMAGE " variable in the github action. 
However I am not sure if it would be the same as  "LIQUIBASE_IMAGE"? do you think it is necessary? @vivid-cpreston

terraform/ecs.tf
```diff
@@ -1357,4 +1357,146 @@ resource "aws_ecs_service" "apisix_gui" {
...
+      memory      = var.server_memory
+      networkMode = "awsvpc"
+      portMappings = [
+        {
+          protocol      = "tcp"
+          containerPort = var.db_port
+          hostPort      = var.db_port
+        }
+      ]
+      environment = [
```
vivid-cpreston, line 1407: Need to include a value for CHANGELOG_FOLDER here. Also need to update wfnews_liquibase resource to include a value for CHANGELOG_FOLDER

vivid-cpreston (changes requested):

terraform/ecs.tf
```diff
@@ -1357,4 +1361,152 @@ resource "aws_ecs_service" "apisix_gui" {
...
+      volumesFrom = []
+    }
+  ])
+}
+
+resource "aws_ecs_service" "notifications_liquibase" {
+  count                             = 1
+  name                              = "notifications-liquibase-service-${var.target_env}"
+  cluster                           = aws_ecs_cluster.wfnews_main.id
+  task_definition                   = aws_ecs_task_definition.wfnews_tasks[wfnews_liquibase].arn
```
vivid-cpreston, line 1481: This should be aws_ecs_task_definition.notifications_liquibase.arn - wfnews_tasks[wfnews_liquibase].arn would be for if the task properties had been added to the tasks map near the top of this file

terraform/ecs.tf
```diff
@@ -1357,4 +1361,152 @@ resource "aws_ecs_service" "apisix_gui" {
...
+      portMappings = [
+        {
+          protocol      = "tcp"
+          containerPort = var.db_port
+          hostPort      = var.db_port
+        }
+      ]
+      environment = [
+        {
+        {
```
vivid-cpreston, line 1413: syntax errror here - you have an extra layer of {}

terraform/variables.tf
```diff
@@ -351,12 +357,24 @@ variable "liquibase_container_name" {
...
+  type        = string
+}
+
 variable "liquibase_image" {
   description = "Full name of liquibase image"
   type        = string
   default     = ""
 }
 
+variable "notificatinos_liquibase_image" {
```
vivid-cpreston, line 372: I don't think this is needed any more

terraform/ecs.tf
```diff
@@ -1357,4 +1361,152 @@ resource "aws_ecs_service" "apisix_gui" {
 
   tags = local.common_tags
 }
+
+resource "aws_ecs_task_definition" "notifications_liquibase" {
```
vivid-cpreston, line 1365: This is currently in the commented-out section - is this intended?

vivid-cpreston (changes requested):
A few more changes I noticed

vivid-cpreston (approved):

PR #858, opened 2023-07-17
WFNEWS-1178 : Update content-disposition for lightgallery downloads

sukhpalp:


dhlevi (approved):
Should do the trick. An alternative could be something like

```
String fileName = result.getFileName()  +
                             (thumbnail.booleanValue() ? "-thumb" : "") +
                              result.getMimeType() != null ? ("." + result.getMimeType()) : "";
```
Using the mime in place of the file name (in case the filename is wrong)

PR #859, opened 2023-07-17
Lambda zip fix

vivid-cpreston:


dhlevi (approved):

dhlevi:
@vivid-cpreston Looks good, assuming this works we can clean up the bucket stuff that was done as a workaround

PR #869, opened 2023-07-18
Terraform cleanup

vivid-cpreston:


terraform/lambda.tf
```diff
@@ -5,25 +5,25 @@
 # }
 
 
-resource "aws_lambda_layer_version" "wfnews_lambda_layer" {
+data "aws_lambda_layer_version" "wfnews_lambda_layer" {
```
dhlevi, line 8: I tried data before and it choked until I added resources, hopefully not the case with the revisions!

dhlevi (approved):

PR #882, opened 2023-07-19
WFNEWS-1265

dhlevi:
Styling for RoF Pages


client/wfnews-war/src/main/angular/src/app/app.component.scss
```diff
@@ -240,6 +240,15 @@
     display: flex !important;
     margin-left: auto;
   }
+  ::ng-deep .wfnews-application {
+    overflow: hidden;
+    position: absolute;
+    top: 0;
```
yzlucas, line 246: we can always just use "inset: 0;" for all four position proerties.

yzlucas (approved):

PR #906, opened 2023-07-21
chips  component upgrade

yzlucas:
fyi, we dont use mat-chip anywhere in the component


dhlevi:
Yeah took a look through the code to be sure and we aren't using chips at the moment. I'd say the better thing to do here is to remove it entirely

yzlucas:
> Yeah took a look through the code to be sure and we aren't using chips at the moment. I'd say the better thing to do here is to remove it entirely

For sure, MatChips has been removed

PR #945, opened 2023-07-25
wfnews-1317

yzlucas:


dhlevi (approved):
Looks like a good chunk of the todo's are cleaned up, which is great. I'll need to test it more thoroughly to verify the changes as there's a lot going on here, but overall it looks good

PR #959, opened 2023-07-26
adding systemProperties for notifications-api

yzlucas:


dhlevi (approved):

dhlevi:
If the environment variable names have changed at all, remember that we may need to update them in the code where they are used. Might be worth some investigation to make sure what's defined in the terraform/action environment vars matches what the API is expecting to use

vivid-cpreston:
I checked the variable names in ecs.tf against the ones in application.properties and application-secrets.properties, and they match up

vivid-cpreston (approved):

PR #967, opened 2023-07-27
WFNEWS-1305 Set default link when evac order has no URL

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.html
```diff
@@ -29,7 +29,8 @@
       <h4>Active Evacuation Orders and Alerts</h4>
       <div class="evac-list" *ngFor="let evac of evacOrders">
         <mat-icon style="text-decoration: none;">error</mat-icon>
-        <a class="label" [href]="evac.uri ? evac.uri : ''" target="_blank" rel="noopener">{{evac.eventName}}</a>
+        <a *ngIf="evac.uri" class="label" [href]="evac.uri ? evac.uri : ''" target="_blank" rel="noopener">{{evac.eventName}}</a>
```
dhlevi, line 32: You could probably just remove the ngIf and extra anchor here:

```
<a class="label" [href]="evac.uri ? evac.uri : 'https://www.emergencyinfobc.gov.bc.ca/current-wildfire-information-2023'" target="_blank" rel="noopener">{{evac.eventName}}</a>
```

Otherwise the change is fine. For the same change on the main repo, we'll likely want this URL to be configurable

dhlevi (approved):

PR #995, opened 2023-08-01
WFNEWS-1292 Use Timestamp JDBC type for Attachment timestamps

ssylver93:


dhlevi (approved):
Change is fine, just curious why were updating these from Dates to Timestamps?

PR #1003, opened 2023-08-01
WFNEWS-1305 Default URL for evac orders/alerts

ssylver93:


yzlucas (commented):
looks good, We can just declare "defaultEvacURL" as string if we are not passing value to this component

yzlucas (approved):

dhlevi (approved):

PR #1016, opened 2023-08-08
Add empty deploy-patch actions to allow enable runs in branches

vivid-cpreston:


vivid-cpreston:
Add empty (non-functional) deploy patch actions to main branch, to enable running in other branches

PR #1018, opened 2023-08-09
ROf styling tweak

ssylver93:


yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/complex-question-page/rof-complex-question-page.component.scss
```diff
@@ -2,7 +2,7 @@
 
 ::ng-deep.mat-button-toggle-label-content {
   line-height: 35px !important;
-  padding-bottom: 1.5rem !important
+  padding-bottom: 3% !important;
```
dhlevi, line 5: Are we sure a percentage is good here? Remember that mean it'll grow with screen size differences

ssylver93: Changed to fixed size

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.scss
```diff
@@ -13,7 +13,7 @@
         margin-bottom: 24px;
         border: none;
         border-radius: 8px;
-        height: 60px;
+        height: 80px;
```
yzlucas, line 16: code conflict here, We may want to use 81px which is same as complex-questions buttons. 
Or use 80px on both.

yzlucas (changes requested):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.scss
ssylver93: Changed to 81px

yzlucas (approved):

PR #1021, opened 2023-08-10
Wfnews 1253 location of fire

yzlucas:


client/wfnews-war/src/main/angular/package.json
```diff
@@ -61,6 +61,7 @@
     "@ngrx/store-devtools": "15.4.0",
     "@qqnluaq/smk": "=1.0.16000",
     "@swimlane/ngx-charts": "20.1.0",
+    "@turf/turf": "^6.5.0",
```
dhlevi, line 64: Turf is already a part of SMK so you shouldn't need to import it. Also, you can import the specific modules you need, rather than the whole library

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/location-page/rof-location-page.component.html
```diff
@@ -1,6 +1,14 @@
 <div class="page-content">
   <div class="rof-text-primary-header">{{ title }}</div>
   <div class="rof-text-secondary-header">{{ message }}</div>
+  <div>
```
dhlevi, line 4: Not sure if we need to bring in SMK for the location map as it doesn't have any actual map functionality? Might make more sense to just use a basic leaflet implementation (see the incident details header map)

yzlucas: Im not sure if leaflet can drag the map with arrow and my location ?  Also the appearance of scale tool in leaflet are diferent as SMK. 
I got the refrence from PM and thought there was a reason to use SMK.

dhlevi: I'm not saying "dont use SMK", just wondering if it's too heavy-weight for the page. If we want specific functionality that it provides though, no big deal to utilize it.

dhlevi (approved):

PR #1032, opened 2023-08-14
ROF: question_mark icon does not exist, Use Next button when there is progress bar

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/complex-question-page/rof-complex-question-page.component.html
```diff
@@ -13,7 +13,7 @@
             <img [alt]="button.label" class="btn-icon" *ngIf="button.customIcon" [src]="button.customIcon" height="20" width="20" />
             <mat-icon *ngIf="button.icon">{{ button.icon }}</mat-icon>{{ button.label }}</button>
         </div>
-        <mat-button-toggle style="width: 100%; box-sizing: border-box;" *ngIf="allowIDontKnowButton" value="skip" (click)="onValChange(null)"><mat-icon>question_mark</mat-icon>I'm not sure</mat-button-toggle>
+        <mat-button-toggle style="width: 100%; box-sizing: border-box;" *ngIf="allowIDontKnowButton" value="skip" (click)="onValChange(null)"><span style="color:#548ADB; margin-right: 8px;">?</span>I'm not sure</mat-button-toggle>
```
dhlevi, line 16: We should try to keep CSS in the stylesheet where possible so we don't copy-paste identical stuff and have to change it in multiple locations.

For the question mark icon, if it's no longer available, do we have an SVG we can grab via figma?

yzlucas: Will move them. 
There is no SVG, It just use text "?" in Figma.

dhlevi: ok

PR #1034, opened 2023-08-15
WFNEWS-1151 Enable push notitifications

ssylver93:


dhlevi (approved):
Looks good but definitely worth a review by @vivid-cpreston  as I didn't set this up with the old app

PR #1035, opened 2023-08-16
Wfnews 1252

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/photo-page/rof-photo-page.component.ts
```diff
@@ -9,11 +10,63 @@ import { ReportOfFire } from "../reportOfFireModel";
...
     super()
   }
 
   initialize (data: any, index: number, reportOfFire: ReportOfFire) {
     super.initialize(data, index, reportOfFire)
   }
+
+  async takePhoto(){
+    this.image = await Camera.getPhoto({
+      quality: 100,
```
dhlevi, line 31: Should any of these hardcoded configs be moved into an external config, so we can adjust them?

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/photo-page/rof-photo-page.component.ts
yzlucas: I dont think we would potentially change them?
quality 100 is the highest value (0 -100) 
allowEditing false is not giving an option for user to crop the photo after taking it. 
Other values are pretty standard.

PR #1051, opened 2023-08-22
disable addFromCameraRoll when condition applys

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/photo-page/rof-photo-page.component.ts
```diff
@@ -2,6 +2,7 @@ import { Component, ChangeDetectionStrategy, ChangeDetectorRef, ElementRef } fro
 import { RoFPage } from "../rofPage";
 import { ReportOfFire } from "../reportOfFireModel";
 import { Camera, CameraResultType, CameraSource } from '@capacitor/camera';
+import { CommonUtilityService } from "../../../../../src/app/services/common-utility.service";
```
dhlevi, line 5: I think it's time we start using absolute paths for this kind of thing, to hopefully prevent some build errors.

Add a section to the tsconfig for paths

```
"paths": {
    "@services/*": ["./src/app/services/*"]
  }
```
Or whatever the path is. Then you can use an import like:
`import { CommonUtilityService } from '@services'`

yzlucas: ok, done!

client/wfnews-war/src/main/angular/tsconfig.json
```diff
@@ -6,7 +6,10 @@
     "typeRoots": [
       "node_modules/@types"
     ],
-    "baseUrl": "./",
+    "baseUrl": "./src",
```
dhlevi, line 9: You shouldn't have to change the baseUrl, but unless you get any build weirdness it shouldn't matter

dhlevi (approved):

PR #1077, opened 2023-09-05
change tileLayer for satellite view

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
```diff
@@ -31,8 +31,8 @@ export class RoFReviewPage extends RoFPage implements AfterViewInit{
       scrollWheelZoom: false
     }).setView(location, 9)
     // configure map data
-    L.tileLayer('http://{s}.google.com/vt/lyrs=s&x={x}&y={y}&z={z}', {
-      maxZoom: 20,
+    L.tileLayer('https://server.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}', {
```
dhlevi, line 34: Might be better to move this into a configuration for easier changing in the future, but not a big deal for now

dhlevi (approved):

PR #1081, opened 2023-09-08
Compass non-compliance WFNEWS-1438, 1443, 1444

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.html
```diff
@@ -6,10 +6,10 @@
         <img alt="Compass Hand" [src]="compassHandUrl"  class="compass-hand-image"/>
     </div>
     <div class="compass-heading" id="compass-heading">{{ heading }}</div>
-    <div class="location">{{ currentLat }},{{ currentLong }}</div>
+    <div class="location" id="location">{{ currentLat }},{{ currentLong }}</div>
     <div class="rof-button-container">
       <button class="rof-button" mat-flat-button (click)="previous()">Previous</button>
-        <button class="rof-button-primary" mat-flat-button (click)="next()">Continue</button>
+        <button class="rof-button-primary" mat-flat-button (click)="next(); confirmHeading()">Continue</button>
```
dhlevi, line 12: In this case, you can likely move the call to `next()` directly into the `confirmHeading()`. Cleaner that way. Make sure to try/catch the `confirmHeading()` so it doesn't puke and fail the next

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -18,14 +18,14 @@ interface DeviceOrientationEventiOS extends DeviceOrientationEvent {
 export class RoFCompassPage extends RoFPage implements OnInit {
   public compassFaceUrl: string
   public compassHandUrl: string
-  public compassHeading: number;
-  public currentLat: number
-  public currentLong: number
+  public compassHeading: number = 0;
```
dhlevi, line 21: No need to change it, but just as an fyi in typescript if you assign a value like zero, you don't need to define the type as `number`

`public compassHeading = 0;` will auto-derive the type.

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -98,16 +98,23 @@ handler(e) {
...
 async useMyCurrentLocation() {
 
   const location = await this.commonUtilityService.getCurrentLocationPromise()
-  if (location) {
+  if (location) {   
     this.currentLat = Number(location.coords.latitude);
     this.currentLong = Number(location.coords.longitude);
   }
+
+  document.getElementById("location").innerText = this.currentLat + "," + this.currentLong;
```
dhlevi, line 112: Always good to wrap getElement calls by a null check, in case the element doesn't exist. Even if you know it will... sometimes weird bugs pop up this way

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
ssylver93: Thanks I updated that now!

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
ssylver93: Ok good to know

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.html
ssylver93: Sounds good, updated accordingly

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -93,21 +95,50 @@ handler(e) {
...
+  try{
+    this.reportOfFire.compassHeading = this.compassHeading;
+    this.next();
+  } catch(err){
+    console.error('Could not confirm heading', err)
+  }
+
+}
+
+formatDDM(decimal: number){
```
dhlevi, line 137: A bit late, but worth noting here:

https://github.com/dhlevi/reproj-helper/blob/2cf3c3737a00a6034b8230aeb300a4cca27d6803/src/spatial-utils.ts#L63

See this for an example implementation that handles a number of corner cases

PR #1083, opened 2023-09-08
Wfnews 1298

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
```diff
@@ -46,7 +46,7 @@ export class RoFReviewPage extends RoFPage implements AfterViewInit{
   selectedAnswer(page:any) {
     switch(page.id){
       case 'contact-page' : 
-        return this.reportOfFire.consentToCall ? 'Yes' : 'No';
+        return (this.reportOfFire.consentToCall && this.reportOfFire.consentToCall === 'yes') ? 'Yes' : 'No';
```
dhlevi, line 49: You've hard-coded the values for the consentToCall already, so you could probably just ignore this check and apply the capitalization to the var?

Initialize consentToCall to 'No', then you can ignore the check and do `return this.reportOfFire.consentToCall`

Not important to change and nothing wrong with the implementation, simplifying just prevents some confusion with the statement

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
yzlucas: Good call, thanks! 
Updated

PR #1085, opened 2023-09-11
wfnews 1440

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
```diff
@@ -128,7 +128,15 @@ export class RoFReviewPage extends RoFPage implements AfterViewInit{
       boxZoom: false,
       trackResize: false,
       scrollWheelZoom: false
-    }).setView(location, 9)
+    })
+
+    // Calculate the bounding box
+    const bbox = L.latLngBounds(this.reportOfFire.currentLocation, this.reportOfFire.fireLocation);
```
dhlevi, line 134: Probably worth a validator for the locations to make sure they exist and are valid before the bbox and zoom level changes, just to prevent unexpected exceptions

dhlevi (approved):

PR #1089, opened 2023-09-12
Fix compass distortion, fix bugs for location page and dynamically up…

ssylver93:
…date coordinates

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
@@ -42,12 +42,15 @@ ngOnInit(): void {
 
 async getOrientation() {
   try{
+    let self = this;
     const requestPermission = (DeviceOrientationEvent as unknown as DeviceOrientationEventiOS).requestPermission;
     const iOS = typeof requestPermission === 'function';
     if (iOS) {
     const response = await requestPermission();
         if (response === "granted") {
-          window.addEventListener("deviceorientation", this.handler, true);
+          window.addEventListener("deviceorientation", (function(passedInElement) {
+            return function(e) {self.handler(e, passedInElement); };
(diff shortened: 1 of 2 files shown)


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -42,12 +42,15 @@ ngOnInit(): void {
 
 async getOrientation() {
   try{
+    let self = this;
     const requestPermission = (DeviceOrientationEvent as unknown as DeviceOrientationEventiOS).requestPermission;
     const iOS = typeof requestPermission === 'function';
     if (iOS) {
     const response = await requestPermission();
         if (response === "granted") {
```
dhlevi, line 50: When testing strings, it's always a good idea to ignore case, unless case is important (in this situation, I assume Granted and granted would mean the same thing)

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -98,17 +103,19 @@ handler(e) {
     if (document.getElementById("compass-face-image")) document.getElementById("compass-face-image").style.transform = `rotate(${-compassHeading}deg)`;
     if (document.getElementById("compass-heading")) document.getElementById("compass-heading").innerText = compassHeading.toString() + "° " + cardinalDirection;
 
-    this.reportOfFire.compassHeading = compassHeading;
+    self.reportOfFire.compassHeading = compassHeading;
 
-    this.useMyCurrentLocation();
+    self.useMyCurrentLocation(self);
```
dhlevi, line 108: if `self` is calling `useMyCurrentLocation`, why do you need to pass `self` into it? It should already be accessible via `this`

dhlevi (approved):

PR #1092, opened 2023-09-12
modify actual values of fire size selection

yzlucas:


ssylver93 (approved):

dhlevi (approved):
For the button values, just be sure those "value" changes are also consistent with what we expect to see in the API through to the RoF in WFIM

PR #1100, opened 2023-09-15
WFNEWS-1353 Make sure location event listener does not update when user navigates away from compass page and update location page arrow direction

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -38,6 +39,7 @@ initialize (data: any, index: number, reportOfFire: ReportOfFire) {
   }
 
 ngOnInit(): void {
+  let self = this;
```
dhlevi, line 42: Is this being used? If not we can remove it

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/complex-question-page/rof-complex-question-page.component.ts
```diff
@@ -113,4 +113,10 @@ export class RoFComplexQuestionPage extends RoFPage {
   backToReview() {
     this.reportOfFirePage.edit('review-page')
   }
+
+  previousPage() {
+    if (this.id === 'distance-page') this.reportOfFire.headingDetectionActive = true;
```
dhlevi, line 118: You could probably put this in an override on the distance page itself, rather than adding logic into the complex question page

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.ts
```diff
@@ -69,4 +69,9 @@ export class RoFSimpleQuestionPage extends RoFPage {
   nextPart() {
     this.reportOfFirePage.edit(this.nextId);
   }
+
+  nextPage(){
+    if (this.id === 'callback-page') this.reportOfFire.headingDetectionActive = true;
```
dhlevi, line 74: Same as above, we could probably override on the specific page

dhlevi (approved):

PR #1106, opened 2023-09-16
WFNEWS-1463 Move iOS Motion permission check to title page

ssylver93:
Moving the permission check to title page to set a boolean for the permission globally. Paul says he the compass is static for him on iPhone/iPad so seeing if setting that globally will make a difference. FYI I can't reproduce this issue and works fine for me on both iPhone 14 and iPad.


yzlucas (approved):
looks good. Please remember to resolve the merge conflict

PR #1129, opened 2023-09-28
wfnews-1329 background runner for offline RoF submission

ssylver93:


client/wfnews-war/src/main/angular/src/background.js
```diff
@@ -0,0 +1,111 @@
...
+          // Remove the locally stored data if sync is successful
+          await storage.remove('offlineReportData');
+        }
+      }
+    } catch (error) {
+      console.error('Sync failed:', error);
+    }
+  }
+
+async function submitReportToServer(offlineReport){
```
dhlevi, line 43: These functions exist elsewhere, might be good to keep this in a service class so we dont need to make identical changes in multiple places

client/wfnews-war/src/main/angular/src/background.js
```diff
@@ -0,0 +1,111 @@
...
+    const rofJson = JSON.parse(offlineReport);
+    const resource = rofJson.resource;
+    const image1 = rofJson.image1;
+    const image2 = rofJson.image2;
+    const image3 = rofJson.image3;
+    const rofUrl = "https://wfone-notifications-api.dev.bcwildfireservices.com/rof"
```
dhlevi, line 49: No hard-coded URL's, use an environment specific parameter. This should be in our secrets and applied on build, fetched from the app properties

dhlevi (changes requested):
On the right track. Couple of small things

client/wfnews-war/src/main/angular/src/background.js
dhlevi: Note: I understand they're in the other PR, so we can likely make that change there, merge it in, then update this one to use the shared code

dhlevi:
check the conflict on package-lock. May need to regenerate or drop it an re PR that bit after a merge

client/wfnews-war/src/main/angular/src/app/background/background.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }
+
+async function submitReportToServer(offlineReport){
+    const rofJson = JSON.parse(offlineReport);
+    const resource = rofJson.resource;
+    const image1 = rofJson.image1;
+    const image2 = rofJson.image2;
+    const image3 = rofJson.image3;
+    const rofUrl = "https://wfone-notifications-api.dev.bcwildfireservices.com/rof"
```
dhlevi, line 48: This needs to be parameterized. No hard-coded URLs

ssylver93:
Replaced by PR #1148

PR #1130, opened 2023-09-28
Wfnews 1393

yzlucas:
apply geoJson shows the outline of BC on map when it is offline.
leaflet does not read local store mbTiles files unless the path is running as a server.


dhlevi (approved):
Lets give it a try and see how it looks. I expect there may be some reworking with tiles, zooms, etc.

dhlevi:
Another thought, we could use some clear comments in the code explaining what we're doing for this. A number of complex sections of the application are missing proper commenting

PR #1137, opened 2023-09-29
wfnews-1465 1469 1461 1470 1448

yzlucas:


ssylver93 (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/comment-page/rof-comments-page.component.html
```diff
@@ -5,10 +5,10 @@
     <mat-label>Note</mat-label>
   </div>
   <mat-form-field  class="text-container">
-    <textarea matInput #commentText rows="3" style="height: calc(100vh - 400px);"
```
dhlevi, line 8: Side note, but did we verify that the maxLength in the database is actually 500?

dhlevi (approved):

PR #1145, opened 2023-10-04
ROF Review Step: draw the lines and draw the arrow and lines between …

yzlucas:
…fire location and current location


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
```diff
@@ -200,6 +203,28 @@ export class RoFReviewPage extends RoFPage implements AfterViewInit{
           iconAnchor: [ 14, 14 ]
       } )
     }).addTo(this.map)
+
+    // draw the arrow and lines between fire location and current location
+    let latlngs = Array();
```
dhlevi, line 208: For the code in this block, there should be some defensive checks for nulls. Specifically on the reportOfFire and fireLocation stuff which will throw errors if there are undefined or null parts.

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/review-page/rof-review-page.component.ts
yzlucas: thanks. Updated

PR #1147, opened 2023-10-04
rof-call-page

yzlucas:


dhlevi:
Is this [WFNEWS-1516](https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-1516)? There is no description

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/title-page/rof-title-page.component.ts
```diff
@@ -37,6 +39,7 @@ export class RoFTitlePage extends RoFPage {
 
   openCallPage () {
     // not yet implemented
```
dhlevi, line 41: Does this comment make sense still? Looks like it is implemented?

client/wfnews-war/src/main/angular/src/assets/data/appConfig.json
```diff
@@ -67,6 +67,7 @@
         "AGOLactiveFirest": "https://services6.arcgis.com/ubm4tcTYICKBpist/ArcGIS/rest/services/BCWS_ActiveFires_PublicView/FeatureServer/0/",
         "evacDefaultUrl": "https://www.emergencyinfobc.gov.bc.ca/current-wildfire-information-2023",
         "contactInformation": {
+            "rofPhoneNumber": "1 800 663-5555",
```
dhlevi, line 70: How is this different then the number directly below it? Aside from the obvious, why do we have this in there twice? Might make more sense to split the phone number and cell * number

yzlucas: I believe the below one is for display in "Contact-Us" dialog desktop view only.

dhlevi: Yeah, just wondering if they should be showing the same thing

dhlevi (approved):

PR #1148, opened 2023-10-04
WFNEWS-1329 background runner & handle offline RoF submission

ssylver93:
Refactored to use capawesome API for the background runner as the new capacitor background runner plugin does not seem to support Ionic Storage yet (only CapacitorKV). Moved the common submission functions for offline checking and offline submission the common utility service and report of fire services. Fixed the ionViewDidEnter function pinging the server multiple times per second.


dhlevi:
@ssylver93  Is this a replacement for the other PR?

dhlevi (approved):

ssylver93:
> @ssylver93 Is this a replacement for the other PR?

@dhlevi yes it is, I've closed that PR now

PR #1156, opened 2023-10-06
WFNEWS-1329 offline submission updates

ssylver93:
Parse formData to JSON to be correctly stored in ionic storage, return Promise<boolean> correctly for offline/online check, prevent images being sent as undefined


client/wfnews-war/src/main/angular/src/app/services/report-of-fire-service.ts
```diff
@@ -117,30 +124,39 @@ export class ReportOfFireService {
...
-      });
+      })
     
-        if (response.ok) {
-          // The server successfully processed the report
-          return { success: true, message: 'Report submitted successfully' };
-        } else {
+      if (response.ok) {
+           // The server successfully processed the report
+         return { success: true, message: 'Report submitted successfully' };
```
dhlevi, line 142: Probably should remove the data from local storage after success?

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/services/report-of-fire-service.ts
ssylver93: Thanks, done

PR #1220, opened 2023-10-18
WFNEWS-1324 : Add location icon

sukhpalp:
- Fix some styling
- Add icons
- Implement background/resume check for mobile devices that have to go into settings screen


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.scss
```diff
@@ -437,7 +437,7 @@ mat-panel-title {
     }
 
     .map {
-      height: calc(100% - 128px) !important;
+      height: 90vh !important;
```
dhlevi, line 440: Is this 90vh reliable all the time? If the footer isn't set to 10vh, use a calc from 100 instead

dhlevi (approved):

yzlucas (approved):

PR #1244, opened 2023-10-23
WFNEWS-1324 fixes

sukhpalp:
- Fix map layers
- Fix snack bar visibility on mobile
- Fix location button visibility and behavior


yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="container">
-    <!-- 
+    <!--
```
dhlevi, line 2: We can remove commented out stuff if its not needed

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.html
```diff
@@ -132,11 +132,11 @@ <h2 class="wildfire-count">
             <panel-local-authority *ngIf="selectedPanel === 'local-authorities'"></panel-local-authority>
         </div>
         <div class="map">
-    <!-- 
+    <!--
```
dhlevi, line 135: We can remove commented out stuff if its not needed

dhlevi (approved):

PR #1245, opened 2023-10-23
WFNEWS-1504

dhlevi:
Dashboard min/max size and media tags for mobile/tablet/desktop


dhlevi:
Yeah, agreed. Now that the components are all there and finalized it can be generalized easy enough

PR #1247, opened 2023-10-24
old identify panel should not display in mobile anymore

yzlucas:


yzlucas:
@sukhpalp  Please ignore and close this PR if you already fixed in in your branch. Looks like this change has been reverted in past few days.

PR #1254, opened 2023-10-25
WFNEWS-1043 Full Details - Area Restrictions WIP

ssylver93:


yzlucas (approved):
looks great! some minor thing to notice. 
A merge conlifct about adding duplicate svg files.

client/wfnews-war/src/main/angular/src/app/components/full-details/area-restrictions-full-details/area-restrictions-full-details.component.ts
```diff
@@ -0,0 +1,145 @@
...
+          name = name.replace("Area Restriction", "").trim();
+          this.name = name + " Restricted Area"
+          this.fireCentre = response.features[0].attributes.FIRE_CENTRE_NAME + " Fire Centre";
+          this.issuedDate = convertToDateYear(response.features[0].attributes.ACCESS_STATUS_EFFECTIVE_DATE);
+      }
+   })
+ }
+
+ populateIncident() {
+  // Using these values as a stub until preview page is complete to pass the incident number from.
```
yzlucas, line 125: Based on our conversation, looks like we need to run a spatial query with area restrictions polygon to check overlap incidents. And once we get that, it can calls populateIncident with incident number if there is a result.

yzlucas (approved):

PR #1257, opened 2023-10-26
WFNEWS-1507

dhlevi:
Initial historical widget layout.
Current fetching 20-Year average via a brute force fetch of 20 year history data, but this will likely change with a new endpoint specific for calculating these on the DB side.

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/historical-comparison-widget/historical-comparison-widget.component.html
+++ b/client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/historical-comparison-widget/historical-comparison-widget.component.html
@@ -1,7 +1,32 @@
 <div class="container">
   <div class="widget-header">
-    <span>Historical Comparison</span>
+    <span class="widget-title">Historical Comparison</span>
+    <div class="widget-button-bar">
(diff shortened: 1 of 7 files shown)


client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/historical-comparison-widget/historical-comparison-widget.component.html
```diff
@@ -1,7 +1,32 @@
...
+      <button [disabled]="viewWildfireCounts" (click)="viewWildfireCounts = !viewWildfireCounts" *ngIf="viewWildfireCounts" class="button-selected" mat-flat-button ><img height="24" width="24" src="/assets/images/svg-icons/wildfire-blue.svg" alt="Wildfires" class="button-icon" /><span class="button-label-selected">Wildfires</span></button>
+    </div>
   </div>
   <ng-template #loading><mat-spinner [diameter]="80" class="spinner-position"></mat-spinner></ng-template>
-  <div *ngIf="startupComplete; else loading"></div>
+  <div class="content" *ngIf="startupComplete; else loading">
+    <div class="content-title">{{viewWildfireCounts ? 'Total Wildfires' : 'Hectares Burned Each Year'}}</div>
+    <div class="content-message">{{viewWildfireCounts ? 'One hectare is a unit of area equal to 10,000 square meters or about 2.5 acres of land.' : 'One hectare is a unit of area equal to 10,000 square meters or about 2.5 acres of land.'}}</div>
```
dhlevi, line 14: Note, I am aware the two strings are identical, but that's what's in the mockup. I highly suspect the mockup is incorrect and one of these strings will change to be more Fire Totals specific vs. hectare definitions

PR #1264, opened 2023-10-28
wfnews-1043 Add buffer to polygon and initialize map correctly

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/full-details/area-restrictions-full-details/area-restrictions-full-details.component.ts
```diff
@@ -128,24 +127,36 @@ export class AreaRestrictionsFullDetailsComponent extends FullDetailsComponent i
...
+  });
+
   let xArray: number [] = [];
   let yArray: number [] = [];
 
-  poly.forEach((value) => {
-    xArray.push(value[0]);
-    yArray.push(value[1]);
-  })
+  buffer.geometry.coordinates[0].forEach((value) => {
```
dhlevi, line 156: Turf has a BBOX function that you can use here. `turf.bbox` can take a polygon or multipoly and calculate the bbox from that

Also, try to avoid forEach where we can. It changes scope and is less efficient than using for of

dhlevi (approved):
Some issues with this architecturally, but the code looks good (you may want to simplify the bbox with turf, but that's certainly not a showstopper). I'll bring these up in the main channel

client/wfnews-war/src/main/angular/src/app/components/full-details/area-restrictions-full-details/area-restrictions-full-details.component.ts
ssylver93: I've update now based on your comments, thanks

PR #1265, opened 2023-10-30
Full details update

dhlevi:


dhlevi:
On the Routing, we're passing in params to the full details page rather than a specific page type. So what happens I include two params for conflicting types? `full-details?restrictionID=1` will land me on the Area Restrictions page now, but when we have other types, how do we handle other params? Duplicate Params? Error checks on "not found" and load failures?

ssylver93 (approved):

PR #1272, opened 2023-11-01
wfnews-1582

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/wf-map-container/wf-map-container.component.ts
```diff
@@ -115,15 +115,22 @@ export class WFMapContainerComponent implements OnDestroy, OnChanges {
...
-    let delayTimer;
-    // double click on map does not bring up identify panel, set half second delay to track if zoom in happens before the delay is finished, then cancel the identify
-    delayTimer = setTimeout(() => {
-      this.selectIncidents.emit(identified.featureSet);
-    },500)
+    let delayTimer; // Variable to store the timeout ID
+    
+    // Function to handle the timeout logic
+    function handleTimeout() {
+      self.selectIncidents.emit(identified.featureSet);
```
dhlevi, line 122: just use `this` so it's clear. No need to reference `self`

client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -182,7 +182,7 @@ export class ActiveWildfireMapComponent implements OnInit, AfterViewInit {
   }});
 }
   
-  panToLocation(long, lat) {
+  panToLocation(long, lat, noZoom?) {
```
dhlevi, line 185: Would it be more straightforward to define this as a default? `zoom: number | null = 10`
The boolean usage in panToFeature seems unecessary

dhlevi (approved):

PR #1276, opened 2023-11-03
WFNEWS Missing icons, issued on date for evacs, some css changes

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.scss
```diff
@@ -171,7 +170,7 @@
       box-sizing: border-box;
       align-items: center;
       justify-content: center;
-      min-width: 130px;
+      width: 40dvw;
```
dhlevi, line 173: Careful with using vw/vh/% as this might cause some unexpected behaviours when resizing. Often it's save when included with specific pixel settings in a calc, with min/max width and height settings.

dhlevi (approved):

PR #1289, opened 2023-11-08
Wildfire full details css adjustments

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.html
```diff
@@ -58,19 +58,16 @@ <h4>Active Evacuation Orders and Alerts</h4>
       </div>
     </div>
     <div *ngIf="isMobileView()">
-      <div class="header-content">
-        <div class="item">
-          <button mat-button class="mobile-button">
-            <mat-icon svgIcon="bookmark"></mat-icon>
+      <div class="contact-buttons">
+        <button mat-button class="mobile-buttons">
+            <img class="icon-contact" height="24" width="24" src="/assets/images/svg-icons/bookmark.svg"
```
yzlucas, line 63: we can get rid of height="24" width="24"  from tag now since the same value has been declared in css class.

yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.html
ssylver93: Removed now, thanks for catching that!

PR #1299, opened 2023-11-09
1381 non compliances

yzlucas:
highlight the incident when preview and zoom in


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -114,6 +136,57 @@ export class DraggablePanelComponent implements OnInit, OnChanges {
...
+      iconAnchor: [12, 12],
+      popupAnchor: [10, 0],
+      shadowSize: [0, 0],
+      className: 'animated-icon'
+    });
+    this.marker = L.marker([Number(incident.latitude), Number(incident.longitude)],{icon: pointerIcon})
+    this.marker.on('add',function(){
+        const icon: any = document.querySelector('.animated-icon')
+        incident
+        if (incident.fireOfNoteInd) {
```
ssylver93, line 160: There is a common function in index.ts called **setDisplayColor** that should work for this

ssylver93 (approved):

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
yzlucas: thanks, I made the change

PR #1305, opened 2023-11-10
Wfnews 1608

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.scss
```diff
@@ -1,6 +1,6 @@
 .draggable-panel {
   background-color: #ffffff;
-  height: 50vh;
+  height: 40vh;
```
dhlevi, line 3: Watch out using VH without a max bounds, it rarely works as well as you'd like. Not sure if we have one for the draggable panel, so likely not an issue in this case, but I've noticed a lot of VH/VW/% usage in the css without bounds and it often causes weirdness on responsive scaling

dhlevi (approved):

PR #1331, opened 2023-11-16
WFNEWS-1596 Send visible flame array to notifications api

ssylver93:


dhlevi (approved):

dhlevi:
Change is good. Do we actually track VisibleFlame in the API?

PR #1334, opened 2023-11-16
Wfnews 1634

yzlucas:
the changes of bans-full-detail.component.html looks overwhelming but the only different is I added a ngIf at top to avoid "undefined" error


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -293,7 +299,7 @@ export class DraggablePanelComponent implements OnInit, OnChanges {
     }
   }
 
-  zoomIn(level?: number) {
+  zoomIn(level?: number, polygon?: boolean) {
```
dhlevi, line 302: Notice that you call it above as `zoomIn(null, true)`. You could simplify that by reversing the inputs
`zoomIn(polygon?: boolean, level?: number)` then change the calls to `zoomIn(true)`

Not really worth changing, just a thought on keeping the inputs simple

dhlevi (approved):

PR #1352, opened 2023-11-20
Add support for youtube list view

dhlevi:
Add client side endpoint for fetching list of youtube video's via the youtube API.
Secret already added in secrets.


dhlevi:
Merged. Need to verify the secrets and client send controller for the next phase on the dashboard widget.

PR #1359, opened 2023-11-22
notification url config

yzlucas:


ssylver93 (changes requested):
Looks good, but I think the new variable also needs to be added in appConfig.mobile.json to make it available in the standalone app

yzlucas:
> Looks good, but I think the new variable also needs to be added in appConfig.mobile.json to make it available in the standalone app

Thanks, updated!

ssylver93 (approved):

PR #1377, opened 2023-11-27
WFNEWS-1581

dhlevi:
Delayed identify state handlers for double-click events. May need to adjust timeout delay


dhlevi:
Hm @yzlucas Still testing this, looks like the additional handlers cause identify to execute twice which is not handling the default panel correctly. If I can't find a suitable correction for that we'll close this PR. Please nobody merge this one until verified!

PR #1396, opened 2023-12-01
WFNEWS-1525 saved location full details

ssylver93:
Have some styling issues to fix and the bcws point ID /weather endpoint is throwing a different error now so I need to look more into that. But the bulk of the ticket is here


client/wfnews-war/src/main/angular/src/app/components/saved/saved-location-full-details/saved-location-full-details.component.ts
```diff
@@ -0,0 +1,253 @@
...
+    await this.fetchNearbyWildfires()
+    this.getFireCentre(this.location)
+  }
+
+  async fetchSavedLocation() {
+    if (this.params && this.params['name']) {
+      await this.notificationService.getUserNotificationPreferences().then(response => {
+        if (response.notifications) {
+          for (const item of response.notifications) {
+            if (item?.notificationName === this.params['name'])
```
yzlucas, line 62: I think notification name is not a primary key so just using the notification_name as param may be problematic. 
Could be more accurate to also use the lat and long as params? So use name + long + lat should match with one of the notifcations.

yzlucas (changes requested):
let's combine the use of notificationName and coordinates to get full detail page

client/wfnews-war/src/main/angular/src/app/components/saved/saved-location-full-details/saved-location-full-details.component.ts
ssylver93: Done, thank you

PR #1421, opened 2023-12-05
Wfnews 1526

yzlucas:


dhlevi (approved):
Remember we can clear out comments we're not using anymore. No need to keep commented code in there.

Lets get testing the notifications!

PR #1469, opened 2023-12-11
notificatio pop up messge update

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/saved/add-saved-location/add-saved-location.component.ts
```diff
@@ -87,7 +87,7 @@ export class AddSavedLocationComponent implements OnInit{
   ngOnInit(): void {
     this.useMyCurrentLocation();
     this.route.queryParams.subscribe(params => {
-      if (params){
+      if (params && params.location){
```
dhlevi, line 90: Could change to `if (params?.location) {` for simplicity

dhlevi (approved):

PR #1502, opened 2023-12-14
In app notifications

yzlucas:


yzlucas:
merge in for now to test the result. Will revert this back if issue occurs

PR #1510, opened 2023-12-15
WFNEWS-1780 fire centre/danger rating for saved locations on iOS

ssylver93:
Could not seem to get past unknown HttpErrorResponse with status code 0 in iOS for fetching from open maps API (possibly server-side issue?) so changing to fetch from AGOL for fire centre and pointID for danger rating.


dhlevi:
Do we need to add `AGOL_FIRE_ZONE_BOUNDARIES` to the terraform scripts as well? Or is it already there?

ssylver93:
> Do we need to add `AGOL_FIRE_ZONE_BOUNDARIES` to the terraform scripts as well? Or is it already there?

@dhlevi I put it as a Repository variable as there is no "test" version of the URL. The rest of the repository variables seem to just be configured in appConfig.jsp, config.jsp, ios.yml and android.yml so I was following that pattern

yzlucas:
I think openMaps API really doesn't like IOS, I have same issue on identify the evacs, danger ratings, bans etc (all from openMaps) . We need to find a way to get around with it otherwise the map wont work well on IOS application

dhlevi:
@yzlucas Is this a permission thing? Why would iOS be problematic unless we're using http or something requiring a specific permissions?

ssylver93:
Closing PR & replacing with https://github.com/bcgov/nr-bcws-wfnews/pull/1523

PR #1511, opened 2023-12-15
WFNEWS-1642 refesh tile cache on initialization

ssylver93:


yzlucas:
Seems a lot of linter clean up? 
is this the main change for this PR? : 

function changeBasemapCacheToken() {
    baseMapCacheToken = Math.trunc(Math.random() * 1e10)
}

ssylver93:
> Seems a lot of linter clean up? is this the main change for this PR? :
> 
> function changeBasemapCacheToken() { baseMapCacheToken = Math.trunc(Math.random() * 1e10) }

@yzlucas that seems to be the main part of the changes for sure. I was pretty much trying to copy what was done for the same change in PM which you can see here https://apps.nrs.gov.bc.ca/int/stash/projects/WFONE/repos/wfone-public-mobile-war/pull-requests/702/overview

dhlevi:
changeBasemapCacheToken function looks like it's calling itself. Maybe rename it in the class function so the difference is obvious?

client/wfnews-war/src/main/angular/src/app/services/wf-map.service.ts
```diff
@@ -25,31 +25,35 @@ export class WFMapService {
...
         this.identifyCallback = cb;
     }
 
-    setIdentifyDoneCallback( cb ) {
+    setIdentifyDoneCallback(cb) {
         this.identifyDoneCallback = cb;
     }
 
-    createSMK( option: any ) {
+    changeBasemapCacheToken() {
```
dhlevi, line 44: At this spot.
```changeBasemapCacheToken() {
        changeBasemapCacheToken()
    }```
looks a bit weird

ssylver93: That's fair, updated now

ssylver93:
> changeBasemapCacheToken function looks like it's calling itself. Maybe rename it in the class function so the difference is obvious?

Changed to changeCacheToken to make it more obvious, thanks!

PR #1546, opened 2024-01-24
wfnews-1954

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.ts
```diff
@@ -422,6 +422,15 @@ this.inputAutoComplete.openPanel();
           if (params['featureType'] === 'BCWS_ActiveFires_PublicView') {
             //wildfire notification
             try {
+              const today = new Date();
```
dhlevi, line 425: Already merged, but this code already exists elsewhere in the UI. Ideally we're not duplicating it

PR #1550, opened 2024-01-25
WFNEWS-1928 : Switch to dvh units which account for browser UI

sukhpalp:


client/wfnews-war/src/main/angular/src/app/components/active-wildfire-map/active-wildfire-map.component.scss
```diff
@@ -220,7 +220,7 @@ mat-panel-title {
 
   .container {
     height: calc(
-      100vh - env(safe-area-inset-bottom, 20px) - env(safe-area-inset-top, 20px)
+      100dvh - env(safe-area-inset-bottom, 20px) - env(safe-area-inset-top, 20px)
```
dhlevi, line 223: Happy with `dvh`, but we'll have to test to be careful things are spontaneously resizing while working with the map. dvh has a tendency to be weird for sudden resizing and performance.

dhlevi (approved):

PR #1554, opened 2024-01-25
ios app wont open window without https prefix

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/fires-of-note-widget/fires-of-note-widget.component.ts
```diff
@@ -87,13 +91,23 @@ export class FiresOfNoteWidget implements AfterViewInit {
...
 
-  viewIncident(incident) {
-    window.open(
-      '/incidents?fireYear=' +
-        incident.fireYear +
-        '&incidentNumber=' +
-        incident.incidentNumberLabel,
-      '_blank',
-    );
+  async viewIncident(incident) {
```
dhlevi, line 94: Seems like the two methods to navigation should be in their own functions as they're doing very different things.

Wondering if the "iOS" method here is something we would rather use instead of opening a new tab anyway?

yzlucas: That make sense to me, as user can always click back arrow to navigate back to dashboard.

dhlevi: Yeah for sure. Maybe confirm with the client if that's preferred

PR #1556, opened 2024-01-26
WFNEWS-1931

sukhpalp:
This started off as a very simple bug fix, but I noticed that we had some issues on the dashboard page when moving to different device sizes, so I fixed up some alignment and visual quirks.


yzlucas (approved):

yzlucas:
Looks good, please ensure that the changes do not impact other components, which, based on the look of the modifications, should not be the case

PR #1561, opened 2024-01-27
WFNEWS-1731 : Fixing notification map spacing across platforms

sukhpalp:


client/wfnews-war/src/main/angular/src/app/app.component.ts
```diff
@@ -165,6 +157,18 @@ export class AppComponent implements OnDestroy, OnInit, AfterViewInit {
   ) {
   }
 
+  @HostListener('window:orientationchange', ['$event'])
+  onOrientationChange() {
+    console.log('device orientation changed');
```
dhlevi, line 162: We can probably remove the log outs here

PR #1562, opened 2024-01-27
Wfnews 1935

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/search/search-page.component.ts
```diff
@@ -390,20 +390,26 @@ this.userLocationChecked = true;
...
-        ? 1
-        : Number(a.distance || 0) < Number(b.distance || 0)
-          ? -1
-          : a.relevance > b.relevance
-            ? 1
-            : a.relevance < b.relevance
-              ? -1
-              : a.title.localeCompare(b.title),
-    );
+    this.allResultData.sort((a, b) => {
```
dhlevi, line 393: Logical operations looks out of order. I expect you're missing brackets around `(a.type === '' &&  b.type !=== '')`

```
if ((a.type === 'incident' && b.type !== 'incident') ||
          (a.type === 'alert' && b.type !== 'alert') ||
          (a.type === 'order' && b.type !== 'order'))
```

I expect otherwise you'll get a number of false positives

dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/components/search/search-page.component.ts
yzlucas: It should not need brackets as AND (&&) has higher priorites than OR (||).  I did a few tests there and it looks functional as expected (shows incident, then alert, then order, then locations with shortest distance)

dhlevi: OK, just make sure its tested!

dhlevi (approved):

PR #1564, opened 2024-01-29
dont show compass page on lanscape mode

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
```diff
@@ -177,4 +177,10 @@ document.getElementById('location').innerText =
       console.error('Could not confirm heading', err);
     }
   }
+
+  checkIfLandscapeMode() {
+    if (window.innerWidth > window.innerHeight) {
```
dhlevi, line 182: `return  window.innerWidth > window.innerHeight;`

Right now this doesn't return `false`, just `true` or `undefined`.

`if (!e.alpha && !e.webkitCompassHeading && this.checkIfLandscapeMode())`

Make precedence and order of conditions clear. In this case, I think the check can be a `&&` (all conditions must pass for skipping compass)

Also, there's probably a capacitor call we can use to check for the mobile app, but the innerWidth innerHeight compare is probably good enough!

dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/compass-page/rof-compass-page.component.ts
yzlucas: updated. 
I think we still need to use OR this.checkIfLandscapeMode()) for this. 
We dont want to show the compass page on any devices without sensor.  If we use AND here, we will see compass page on Sharon's phone.  
In general, if device has no magnetic sensor, we skip. If device is in landscape mode, we skip. (compass direction is not correct, usually 90 degrees off from correct reading)

yzlucas: The reason I think we cant really do this via CSS is we will skip the compass page and goes to next component (not just hiding) if device is landscape.  I dont think CSS alone will provice boolean value.

yzlucas: Ok, good to acknowledge that! updated

PR #1572, opened 2024-02-01
Wfnews 1631

yzlucas:


client/wfnews-war/src/main/angular/src/app/services/report-of-fire-service.ts
```diff
@@ -93,6 +93,18 @@ formData.append('image3', await this.convertToBase64(image3));
 return;
 }
 
+      const storedOfflineReportData = await this.storage.get('offlineReportData');
```
dhlevi, line 96: awaits from promises should generally have a catch, as right now any exception occurring in the promise may not be handled or caught properly

dhlevi (approved):

PR #1573, opened 2024-02-01
Wfnews 1935 update : mobile map search should always order by best matching keyword instead of distance, after incident types.

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/search/search-page.component.ts
```diff
@@ -70,7 +70,7 @@ export class SearchPageComponent implements OnInit {
       try {
         this.recentData = (
           JSON.parse(localStorage.getItem('recent-search')) as SearchResult[]
-        ).filter((r) => r.type && r.type !== null);
+        ).filter((r) => r?.type && r?.type !== null);
```
dhlevi, line 73: The second condition is not needed, nor are the `?` operators, if all we're doing is checking for null. `r?.type` is the same thing as saying `r?.type !== null` plus includes undefined.

dhlevi (approved):

PR #1574, opened 2024-02-02
wfnews-1940 wfnews-1952 wfnews-1943 wfnews-1428

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/saved/saved-location-full-details/saved-location-full-details.component.ts
```diff
@@ -195,8 +195,12 @@ this.agolService
           .then((bans) => {
             if (bans && bans.features) {
               this.fireBans = [];
-              for (const item of bans.features) {
-                this.fireBans.push(item);
+              for (const item of bans?.features) {
```
dhlevi, line 198: Remove the `?` from this line, update the `if` clause above.

```
if (bans?.features) {
   for (const item of bans.features) {
...
```

client/wfnews-war/src/main/angular/src/app/components/saved/saved.component.ts
```diff
@@ -86,9 +86,18 @@ export class SavedComponent implements OnInit {
         .subscribe((bans) => {
           this.savedLocations[outerIndex].bans = [];
           for (const innerIndex in bans?.features) {
-            const element = bans?.features[innerIndex];
-            this.savedLocations[outerIndex].bans.push(element);
-            this.cdr.markForCheck();
+            if (bans?.features.hasOwnProperty(innerIndex)) {
+              const element = bans?.features[innerIndex];
```
dhlevi, line 90: You can remove the `?` here as the if check above already ensures the value exists

`hasOwnProperty` should not be executed on the object directly, but through the Object prototype

`Object.hasOwn(bans?.features, innerIndex)`

client/wfnews-war/src/main/angular/src/app/components/saved/saved.component.ts
```diff
@@ -86,9 +86,18 @@ export class SavedComponent implements OnInit {
...
+          
+              // Check if the attribute already exists in the savedLocations array
+              const isAttributeAlreadyExists = this.savedLocations[outerIndex].bans.some(existingElement =>
```
dhlevi, line 93: maybe `doesAttributeExist` is a better var name? Current one is bit of an weird mouthful. Could also "not" the attribute here so the `if` statement can be truthy

dhlevi (changes requested):

yzlucas:
Moved attributePresent checker into common-utility.service

ssylver93 (approved):

PR #1581, opened 2024-02-05
wfnews-1986

yzlucas:


yzlucas:
Somehow, we need to decrement the PROT_RA_SYSID by 1 in the AGOL caller 'where=PROT_RA_SYSID=${sysId}' to get the corresponding AreaRestriction's details. It could be that AGOL's database uses zero-based numbering to name their SYSID.

dhlevi (changes requested):
Do not do this. The SysID is a generated primary key and this is not the fix

dhlevi:
Closing as this should not be commited

PR #1588, opened 2024-02-06
Wfnews 1829

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -454,68 +454,74 @@ return 'Unknown';
...
               .then((response) => {
-                if (response?.extent) {
-                  viewer.map.fitBounds(
-                    new L.LatLngBounds(
-                      [response.extent.ymin, response.extent.xmin],
-                      [response.extent.ymax, response.extent.xmax],
-                    ),
-                  );
+                if (
+                  response &&
```
dhlevi, line 470: `response?.features?.length > && response?.features[0].geometry?.rings?.length > 0`

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -454,68 +454,74 @@ return 'Unknown';
...
```
dhlevi, line 493: Same here as above, minimize the If calls. Sonar will bark at this

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -454,68 +454,74 @@ return 'Unknown';
...
               .toPromise()
               .then((response) => {
-                viewer.map.fitBounds(
-                  new L.LatLngBounds(
-                    [response.extent.ymin, response.extent.xmin],
-                    [response.extent.ymax, response.extent.xmax],
-                  ),
-                );
+                if (
+                  response &&
```
dhlevi, line 515: And once again, with feeling

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -769,4 +775,21 @@ type = 'evac-order';
     const formattedDate: string = date.toLocaleDateString('en-US', options);
     return formattedDate;
   }
+
+  fitPolygonToMap(polygonData: number[][], viewerMap: L.Map): L.Polygon {
+    const swappedPolygonData: number[][] = polygonData.map(([latitude, longitude]) => [longitude, latitude]);
+    const polygonShape = L.polygon(
```
dhlevi, line 781: Why are we adding the polygon data to the map? Does it get removed later? I didn't see a removal. If this is just a zoom to bounds, we don't need to add it

dhlevi (changes requested):

PR #1599, opened 2024-02-09
WFNEWS-1927 fix PDF download and ROF gallery images

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-maps-panel/incident-maps-panel.component.ts
```diff
@@ -102,14 +103,37 @@ export class IncidentMapsPanel implements OnInit {
...
     const url = mapLink;
-    const request = this.httpClient.request(
-      new HttpRequest('GET', url, {
-        reportProgress: true,
-        responseType: 'blob',
-      }),
-    );
+    let isMobile = false;
 
+    this.capacitorService.isMobile.then(response => {
```
yzlucas, line 114: I think "isMobile" will still be false because` this.capacitorService.isMobile.then(response => { isMobile = response; })` is asynchronous.

maybe better to use :

`try {
    isMobile = await this.capacitorService.isMobile();
    } catch (error) {}`

yzlucas (changes requested):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-maps-panel/incident-maps-panel.component.ts
ssylver93: Thanks, I added some new changes now

PR #1604, opened 2024-02-12
wfnews-1991

yzlucas:


dhlevi:
No longer needed. We're moving to the AGO layer for now until we can get the cloudfront cache in place. I've put your change here into the WMS version of that layer and merged already.

PR #1612, opened 2024-02-12
wfnews-1956, 1963, 1991, 1958, 1929, 1971, 1967

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/wildfires-list-header/bans-list/bans-list.component.mobile.ts
```diff
@@ -9,7 +9,7 @@ import { ResourcesRoutes } from '@app/utils';
 })
 export class BansListComponentMobile extends BansListComponent {
   navigateToFullDetails(item: any) {
-    if (item.id) {
+    if (item.id !== undefined && item.id !== null) {
```
ssylver93, line 12: Doesn't this change do the same thing as the original code?

ssylver93 (approved):

client/wfnews-war/src/main/angular/src/app/components/wildfires-list-header/bans-list/bans-list.component.mobile.ts
yzlucas: this will exclude the situation when Id =0

PR #1623, opened 2024-02-15
Wfnews 1829 add polygon highlight and pindrop to mobile map identify

yzlucas:


dhlevi (changes requested):
Several comments, some/most are more suggestions or questions than actual tasks or revisions, but I'm marking as "request changes" for now so you can review. If you don't want to make any further changes from the suggestions just put it back to in review

dhlevi:
Also, side note, but noticing your commit history is getting messy. Make sure you're synced with the latest head from the main repo!

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -462,58 +512,80 @@ return 'Unknown';
...
-                  returnGeometry: false,
-                  returnCentroid: false,
-                  returnExtent: true,
+                  returnGeometry: true,
+                },
+              )
+              .toPromise()
+              .then((response) => {
+                if (response?.features?.length > 0 && response?.features[0].geometry?.rings?.length > 0){
+                  const polygonData = response.features[0].geometry.rings[0];
```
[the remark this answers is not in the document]
yzlucas: Thanks! I have made update for all the above comments and created a hull to handle the possiblities of mutypoly for polygon layers.

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -186,14 +197,11 @@ export class DraggablePanelComponent implements OnInit, OnChanges, OnDestroy {
       } else {
         //identify anything other than incident
         if (
-          this.identifyItem.layerId.includes('bans-and-prohibitions') ||
-          this.identifyItem.layerId.includes('evacuation-orders-and-alerts') ||
-          this.identifyItem.layerId.includes('area-restrictions') ||
-          this.identifyItem.layerId.includes('weather-stations')
+          ['bans-and-prohibitions', 'evacuation-orders-and-alerts', 'area-restrictions', 'weather-stations', 'danger-rating'].some(str => this.identifyItem.layerId.includes(str))
```
dhlevi, line 200: You don't need the `some` here. Just `array.includes(layerId)` will return a true/false if the string is in the array (the same as the list with or operators)

`['bans-and-prohibitions', 'evacuation-orders-and-alerts', 'area-restrictions', 'weather-stations', 'danger-rating'].includes(this.identifyItem.layerId)`

dhlevi (changes requested):

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -762,4 +827,40 @@ type = 'evac-order';
       return date.toLocaleDateString('en-US', options);
     } else return '';
   }
+
+  extractPolygonData(response) {
+    const polygonData = [];
+    response.forEach(element => {
```
dhlevi, line 833: Not super important, but typically it's better to use `for(const thing in array)` than `array.forEach` both of performance and scope reasons

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -762,4 +827,40 @@ type = 'evac-order';
       return date.toLocaleDateString('en-US', options);
     } else return '';
   }
+
+  extractPolygonData(response) {
```
dhlevi, line 831: An option here that may be more efficient is to store only the rings coordinates:

```
const points = []
for (const ring of response) {
   ... extract coords
}

then covert to a single polygon convex hull

`turf.convex(points);` or `window['turf'].convex(points);` if we dont have an import here.

You can return the generated convex hull polygon which will let you treat the zoom checks in a simple way

File: client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts (hunk 2 above)
Comment (yzlucas) [reply client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts:200]: The reason I use "some" here is because we have layers end with "bans-and-prohibitions-cat1" or "bans-and-prohibitions-cat2" etc.  So I am checking if the layerID contains the strings I list there

File: client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts [hunk 5]
```diff
@@ -462,58 +500,84 @@ return 'Unknown';
...
+                if (response?.features?.length > 0 && response?.features[0].geometry?.rings?.length > 0){
+                  const polygonData = this.extractPolygonData(response.features[0].geometry.rings);
+                  if (polygonData.length) {
+                    this.fixPolygonToMap(polygonData);
+                  }                
+                }
               });
           }
+        } else if ( // local authorities
+          layerId === 'abms-regional-districts' ||
```
dhlevi, line 570: Optional as always, but can be simplified

```
['abms-regional-districts', 'clab-indian-reserves', 'abms-municipalities'].includes(layerId)
```

dhlevi (approved):

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
yzlucas: sure

PR #1639, opened 2024-02-20
WFNEWS-1927 Download PDFs with capacitor filesystem

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-maps-panel/incident-maps-panel.component.ts
```diff
@@ -108,25 +108,13 @@ export class IncidentMapsPanel implements OnInit {
...
-            params: {responseType: 'blob'}   
-          };
-          response = CapacitorHttp.get(options).then(resp => {
-              this.downloadFile(resp.data, fileName);
-              this.snackbarService.open('PDF downloaded successfully.', 'Close', {
-                duration: 10000,
-                panelClass: 'snackbar-success-v2',
-              });
-          });
+      await this.capacitorService.isMobile.then(async isMobile => {
```
yzlucas, line 113: MInor, but some considerations regarding readbility, may be better to use:

const isMobile = await this.capacitorService.isMobile();
if (isMobile) { 
....
}
else {
....
{

ssylver93: this.capacitorService.isMobile returns a Promise so I think **then** has to be used here. If you look at where it's called elsewhere in the app, that's how it gets called. I removed the async keyword as it's not needed

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-maps-panel/incident-maps-panel.component.ts
```diff
@@ -184,4 +172,28 @@ export class IncidentMapsPanel implements OnInit {
     document.body.removeChild(a);
   }
 
+  async downloadMobileFile(fileName: string, url: string) {    
+    if (!fileName.endsWith('.pdf')) {
+      fileName += '.pdf';
+    }
+    
+    const download = await Filesystem.downloadFile({
```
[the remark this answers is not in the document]
ssylver93: Certainly, I added a try-catch now. Thanks!

yzlucas (approved):

PR #1770, opened 2024-04-11
WFNEWS-2008 : Fix spinner centering

sukhpalp:
This fix seems a little heavy handed, because the change would apply to all images in the map, but the spinner doesn't have a class or id to hook onto.

If you know of a better way, let me know.


dhlevi (approved):
Shouldn't break anything as any other images are either above the map or wrapped in other elements. Something we'd want to comment that it's up for removal once/if SMK adds an ID or class to the spinner though

PR #1773, opened 2024-04-11
wfnews-2083

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/saved/add-saved-location/add-saved-location.component.ts
```diff
@@ -290,11 +290,16 @@ export class AddSavedLocationComponent implements OnInit {
     return this.notificationService
       .getUserNotificationPreferences()
       .then((response) => {
-        if (response.notifications) {
+        if (response?.notifications.length) {
```
ssylver93, line 293: Probably a good idea to do a null check on notifications here before checking length **response?.notifications?.length**

ssylver93 (approved):

PR #1781, opened 2024-04-13
WFNEWS-2051 : Raster map on Android and code formatting

sukhpalp:


ssylver93 (approved):

client/wfnews-war/src/main/angular/src/app/utils/index.ts
```diff
@@ -755,3 +755,5 @@ return;
     return hc.temp != null;
   });
 }
+
+export const isAndroidViaNavigator = () => navigator.platform.includes('Linux') || navigator.platform.includes('Android');
```
yzlucas, line 759: We also have a "isAndroidPlatform" in capacitor-services. Might be able to use that property directly

yzlucas (approved):

PR #1817, opened 2024-04-22
WFNEWS-1993

sukhpalp:


dhlevi (approved):
So far so good. The only missing piece is the terraform scripts for cloudfront

PR #1827, opened 2024-04-23
WFNEWS-1993 : Use variable for service6 URL

sukhpalp:


vivid-cpreston (approved):

client/wfnews-war/src/main/webapp/config.jsp
```diff
@@ -121,9 +121,15 @@
       openmaps = openmaps.substring(0, openmaps.length() - 1); //Strip off trailing slash, if it exists.
     }
 
+    String services6 = EnvironmentVariable.getVariable("SERVICES6_BASE_URL");
```
vivid-cpreston, line 124: I think this needs to be moved above line 69, otherwise will it be defined there?

client/wfnews-war/src/main/angular/src/assets/data/appConfig.json
vivid-cpreston, line 1: I think these should be parameterized in appConfig.mobile.json and set by the iOS/android deployer, since they're no longer the same across all environments

vivid-cpreston (changes requested):

PR #1892, opened 2024-05-08
WFNEWS-2147 : Storybook

sukhpalp:
Mostly autogenerated changes. Please note that the NODE version is being bumped to 18.

### DIFF
--- a/terraform/cloudfront.tf
+++ b/terraform/cloudfront.tf
@@ -915,7 +915,7 @@ resource "aws_cloudfront_distribution" "wfnews_openmaps_cache" {
       }
     }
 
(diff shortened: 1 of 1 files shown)


client/wfnews-war/src/main/angular/android/app/google-services.json
```diff
@@ -0,0 +1,40 @@
+{
```
ssylver93, line 1: Shouldn't this file be removed? We have it in secrets as GOOGLE_SERVICES_JSON_BASE64, if it needs to be updated I think the change should be made in there.

ssylver93 (changes requested):
Just left a comment regarding the google-services.json file, otherwise looks good.

yzlucas (approved):

ssylver93 (approved):

PR #1907, opened 2024-05-14
Wfnews 2097

yzlucas:
storybook coming... 

image


yzlucas:
> Great work, just needs minor adjustments. Please see my other comment.

Thanks! here is the look after CSS change. I will add the missing zoom button in a separate PR

image

PR #1912, opened 2024-05-16
wfnews-2102

yzlucas:
image

image

storybook coming


ssylver93 (approved):

yzlucas:
> These tabs can be pulled out into a component and do not need to be in the page. I would also like to see the component be reusable so we can pass in a collection of tabs and a function for tab change.
> 
> Storybook also missing

Thanks, separated tabs component out. Storybook added

PR #1928, opened 2024-05-31
Storybook icons and tests

sukhpalp:


client/wfnews-war/src/main/angular/src/stories/Iconography.mdx
yzlucas, line 1: I this an auto-generated file?

yzlucas (approved):

PR #1941, opened 2024-06-05
weather desktop identify preview

yzlucas:
image


yzlucas:
> I think the font weight for the "Active Weather Station" is too bold based on the design and the header is not vertically centered.

There was a hard code size for an icon and I added vertical align to the icons now
image

PR #1943, opened 2024-06-05
weather full detail

yzlucas:


yzlucas:
> Code looks great, nice use of components. I think the scrollbar replacement got missed though. You can spawn it out into a new ticket if needed, but it would make sense to implement it in the panel container so it applies to all of the different panels.
> 
> Please include stories.

Not sure about scrollbar replacement, I applied overflow-y: auto to the panel so it will come with scroll bar as the screenshot displayed.

ssylver93 (approved):

yzlucas:
> The scrollbar in the figma is a CSS custom replacement. We would have to hide the browser scrollbar and show that in its place.
> 
> https://ishadeed.com/article/custom-scrollbars-css/

Ah! I got it! Thanks
I didn't notice the difference on mock up but here is the replacement:
image

yzlucas:
Storybook added @sukhpalp

PR #1954, opened 2024-06-12
danger rating font and protected land preview panel zoom in fix

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -600,6 +600,14 @@ return 'Unknown';
             window['turf'].point([long, lat]),
             this.defaultZoomLevel,
           );
+        } else if (layerId.includes('protected-lands-access-restrictions')) {
+          this.identifyItem;
```
ssylver93: How come this gets declared in the function? Won't it already be declared at the class level?

client/wfnews-war/src/main/angular/src/app/components/preview-panels/protected-land-preview/protected-land-preview.component.ts
```diff
@@ -21,5 +30,13 @@ export class ProtectedLandPreviewComponent {
     showPanel('identify-panel-wrapper')
     hidePanel('desktop-preview');
   }
+
+  zoomIn(){
+    const polygonData = this.commonUtilityService.extractPolygonData(this.geometry?.coordinates);
+    if (polygonData?.length) {
```
sukhpalp: if checks for truthy, so you can use just
`if (polygonData) {`

That covers undefined, null, empty string, empty array.

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
sukhpalp: If this is a class variable it should be defined above and this can be removed.

If this is scoped to the function, we should not attach it to the scope.

yzlucas: sorry, typo. was using as debug purpose

yzlucas: sorry, typo. was using as debug purpose

client/wfnews-war/src/main/angular/src/app/components/preview-panels/protected-land-preview/protected-land-preview.component.ts
yzlucas: updated, thanks

sukhpalp (approved):

ssylver93 (approved):

PR #1976, opened 2024-06-18
Component splitting and bug fixes

sukhpalp:
- Split out components, align styling with other tabs
- Fix maps download dates
- Make All Photos button go to Gallery tab
- Fix tab shadow
- Add font fallbacks in some components
- Fix printing formatting

image
image
image


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/fire-size-card/fire-size-card.component.html
```diff
@@ -0,0 +1,14 @@
+<content-card-container class="fire-size-card">
+  <div class="header">
+    <span class="title">Fire Size</span>
+  </div>
+  <div class="content-container">
+    <div class="content">
+      <div class="content-title"><img class="icon" src="assets/images/svg-icons/carbon_ruler.svg" alt="ruler">
+        {{ fireSize.toLocaleString() }} Hectares </div>
```
ssylver93: A safe operator here would be good in case the data is null for whatever reason

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/primary-media-card/primary-media-card.stories.ts
```diff
@@ -0,0 +1,74 @@
+/* eslint-disable max-len */
```
ssylver93: This line could be removed

ssylver93 (changes requested):
Looks good, just a couple of comments

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/fire-size-card/fire-size-card.component.html
sukhpalp: fireSize is defaulted so that should never be the case, but I can add one.

ssylver93 (approved):

PR #1981, opened 2024-06-19
Response update card keep shaking when there is only one line of text

yzlucas:
Using AfterContentChecked can indeed lead to performance issues if it is triggered frequently, as it is called after every change detection cycle. This also means this.cdr.detectChanges() gets called too frequently.

Paul Pilon
  3:04 PM
[@Lucas Li](https://vividsolutionsinc.slack.com/team/U01D5LHR0DR)
 If you try to access response section this page, does it look jittery?  Esu noticed that, and I see the same thing on my computer.  Seems to only be ones that only have one line of response update while everything else is default.
https://wfnews-client.test.bcwildfireservices.com/incidents?fireYear=2023&incidentNumber=V65581&source=list

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/response-update-card/response-update-card.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/response-update-card/response-update-card.component.ts
@@ -19,7 +19,7 @@ export class ResponseUpdateCardComponent implements AfterViewInit {
 
   ngAfterViewInit() {
     const size = this.contentElementRef?.nativeElement?.offsetHeight;
-    this.useColumns = size >= (this.columnTrigger * this.columnSize);
+    this.useColumns = size > (this.columnTrigger * this.columnSize);
     if (this.useColumns) {
       this.cdr.detectChanges();
     }
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/response-update-card/response-update-card.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/response-update-card/response-update-card.component.ts
@@ -1,11 +1,11 @@
-import {AfterViewInit, ChangeDetectorRef, Component, ElementRef, Input, ViewChild } from '@angular/core';
+import {AfterContentChecked, ChangeDetectorRef, Component, ElementRef, Input, ViewChild } from '@angular/core';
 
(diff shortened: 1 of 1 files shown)


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/response-update-card/response-update-card.component.ts
```diff
@@ -1,11 +1,11 @@
-import { AfterContentChecked, ChangeDetectorRef, Component, ElementRef, Input, ViewChild } from '@angular/core';
+import {AfterViewInit, ChangeDetectorRef, Component, ElementRef, Input, ViewChild } from '@angular/core';
 
 @Component({
   selector: 'response-update-card',
   templateUrl: './response-update-card.component.html',
   styleUrls: ['./response-update-card.component.scss']
 })
-export class ResponseUpdateCardComponent implements AfterContentChecked {
+export class ResponseUpdateCardComponent implements AfterViewInit {
```
sukhpalp: Would AfterContentInit be the lifecycle hook we want? It happens once but only once the DOM is initiated and works better for things like ng-content

sukhpalp: After looking into the lifecycles a bit more, I don't think the code would get the DOM sizes without the DOM being fully drawn, which is why AfterContentChecked is used. We likely would need to update the check instead.

sukhpalp: I tried to get too fancy and detect lines of text, but it might just be better to use a fixed threshold like this.

`
  heightThreshold = 200;

  constructor(private cdr: ChangeDetectorRef) { }

  ngAfterContentChecked() {
    const size = this.contentElementRef?.nativeElement?.offsetHeight;
    const previousUseColumns = this.useColumns;
    this.useColumns = size > this.heightThreshold;
    if (this.useColumns !== previousUseColumns) {
      this.cdr.detectChanges();
    }
  }  
}

`

sukhpalp (changes requested):
Changing the lifecycle will break the column check and not work for large response updates, I provided a sample of something that may work.

yzlucas:
> Changing the lifecycle will break the column check and not work for large response updates, I provided a sample of something that may work.

Thanks, it make a lot of sense. Updated

ssylver93 (approved):

PR #1982, opened 2024-06-20
WFNEWS-2206 : Only call for required bans and make tiles larger

sukhpalp:
Not sure if this totally solves the problem, but it should help alleviate the number of calls.

Might want to hide whitespace.


yzlucas (approved):
This might reduce the number of calls at same time. 
My another thought is if we should also add 0.5s timeout between each adding bans' layer?

sukhpalp:
> This might reduce the number of calls at same time. My another thought is if we should also add 0.5s timeout between each adding bans' layer?

The issue seems to be moreso the call between the tiles because its based on a referer, so I don't know if adding it between the calls would help, but if this doesn't work, we can try it.

PR #1986, opened 2024-06-21
WFNEWS-2237 Await evac response from AGOL

ssylver93:


client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/evac-orders-details-panel/evac-orders-details-panel.component.ts
```diff
@@ -107,11 +107,11 @@ export class EvacOrdersDetailsPanel implements OnInit {
...
   }
 
-  getEvacOrders() {
+  async getEvacOrders() {
     if (this.incident.geometry.x && this.incident.geometry.y) {
-      this.agolService
+      await this.agolService
         .getEvacOrders(null, this.incident.geometry)
-        .subscribe((response) => {
+        .toPromise().then((response) => {
```
sukhpalp: Let's keep this consistent. We are using `await` but then using `then`. You can do that assignment to response by doing something like 

`const response = await this.agolService.().getEvacOrders(null, this.incident.geometry).toPromise()`.

We should also make sure the call is being wrapped in a try/catch so we can gracefully handle the error.

sukhpalp (changes requested):

client/wfnews-war/src/main/angular/src/app/components/admin-incident-form/evac-orders-details-panel/evac-orders-details-panel.component.ts
ssylver93: Thanks, I've updated now

sukhpalp (approved):

PR #2012, opened 2024-07-10
Show incidentCauseDetail on desktop public side if it exists

ssylver93:
and show default disclaimer if it is null. I think this functionality was removed when the bug for defaults was fixed. It works on the mobile version already.

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/suspected-cause-card/suspected-cause-card.component.html
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/suspected-cause-card/suspected-cause-card.component.html
@@ -7,7 +7,7 @@
       <div class="content-title"><img class="icon" [src]="getCauseIcon()" alt="authorities">
         {{ getCauseLabel() }}</div>
       <div class="content-list">
-        <span class="content-text">{{ getCauseDescription() }}</span>
+        <span class="content-text">{{ incident.incidentCauseDetail ? incident.incidentCauseDetail: getCauseDescription() }}</span>
       </div>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


sukhpalp (commented):
Not sure I understand the change to the component. Why not check the value in an ngIf on the incident-info-panel level?

ssylver93:
> Not sure I understand the change to the component. Why not check the value in an ngIf on the incident-info-panel level?

@sukhpalp right now it is just the `generalIncidentCauseCatId` being passed to the SuspectedCauseCardComponent. I'm changing it to pass the full incident so that the `incidentCauseDetail` can be used if it's not null, and if not `generalIncidentCauseCatId` can be used to set a default. I'm not sure what an ngIf would look like at the higher level as there are two different values that possibly need to be used?

sukhpalp:
> > Not sure I understand the change to the component. Why not check the value in an ngIf on the incident-info-panel level?
> 
> @sukhpalp right now it is just the `generalIncidentCauseCatId` being passed to the SuspectedCauseCardComponent. I'm changing it to pass the full incident so that the `incidentCauseDetail` can be used if it's not null, and if not `generalIncidentCauseCatId` can be used to set a default. I'm not sure what an ngIf would look like at the higher level as there are two different values that possibly need to be used?

I see, that makes sense. I think it would be a bit more clear if we determined the text to display in the TS. I'll leave a comment in the code to show what I mean.

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-tabs/cards/suspected-cause-card/suspected-cause-card.component.html
```diff
@@ -7,7 +7,7 @@
       <div class="content-title"><img class="icon" [src]="getCauseIcon()" alt="authorities">
         {{ getCauseLabel() }}</div>
       <div class="content-list">
-        <span class="content-text">{{ getCauseDescription() }}</span>
+        <span class="content-text">{{ incident.incidentCauseDetail ? incident.incidentCauseDetail: getCauseDescription() }}</span>
```
sukhpalp: Can we make this into a function in the TS code. We aren't working with strings here, so it keeps it cleaner to have the logic in one place.

sukhpalp (approved):

PR #2014, opened 2024-07-11
WFNEWS-2294

ssylver93:
Screenshots for making the border of fire bans thicker & darker for desktop full details;
image
image
image
image
image


client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
```diff
@@ -82,6 +83,7 @@ export class DraggablePanelComponent implements OnInit, OnChanges, OnDestroy {
   displayItemTitle = displayItemTitle;
   addMarker = addMarker;
   getStageOfControlDescription = getStageOfControlDescription;
+  displayDangerRatingDes = displayDangerRatingDes;
```
sukhpalp: is Des short for Description? If so can we rename it to use the whole word?

sukhpalp (approved):
One small change, but the rest looks good. Nicely done.

client/wfnews-war/src/main/angular/src/app/components/draggable-panel/draggable-panel.component.ts
ssylver93: Sure, updated that now

PR #2029, opened 2024-07-15
Create a cache behavior for statistics

sukhpalp:


dhlevi (approved):
Looks good. Not sure if we want to up the min ttl in case it's a bit short for the longer queries

vivid-cpreston (approved):
Looks good!

PR #2030, opened 2024-07-16
fix forloop for extractPolygonData

yzlucas:


yzlucas:
merge in for now to test in app, can be reverted if something comes up

PR #2032, opened 2024-07-16
Pipeline docker branch fix

sukhpalp:


vivid-cpreston (approved):

.github/workflows/deploy-prod.yml
```diff
@@ -1,7 +1,7 @@
 name: Deploy Prod
 
 env:
-  IMAGE_TAG: ${{ inputs.IS_HOTFIX == 'true' && format('hotfix-{0}-{1}', vars.PATCH_BRANCH, vars.HOTFIX_TAG) || github.ref_name }}
```
vivid-cpreston: This is the one spot where we don't want to change - having the ref_name here lets us use the unique tag for that release

vivid-cpreston (changes requested):

vivid-cpreston (approved):

PR #2062, opened 2024-07-26
Add caching to weather endpoint

sukhpalp:


dhlevi (approved):
Looks good. What kind of cache metrics are we seeing for this? How long does it maintain?

sukhpalp:
> Looks good. What kind of cache metrics are we seeing for this? How long does it maintain?

I'm trying to build a dashboard so we can monitor the impact. I have been having a tricky time getting this running locally, and even then the lack of traffic would make it hard to see meaningful data.

With a dashboard made, I am hoping to monitor the DB usage over time.

PR #2064, opened 2024-07-26
adding evac order layer and fire perimeter layer to incident detail page

yzlucas:
image

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.ts
@@ -73,6 +73,41 @@ export class IncidentHeaderPanelComponent implements AfterViewInit, OnInit {
 
   private map: any;
 
+  private perimeterLayer = esri.featureLayer({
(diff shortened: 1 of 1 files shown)


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.ts
```diff
@@ -451,6 +451,44 @@ export class IncidentHeaderPanelComponent implements AfterViewInit, OnInit {
           fillColor: colorToDisplay,
         }).addTo(this.map);
       }
+
+      esri.featureLayer({
```
sukhpalp: can we put these layers in a const up top and refer to that to avoid code duplication?

sukhpalp (changes requested):
One small change to keep the code DRY

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/incident-info-header/incident-header-panel.component.ts
yzlucas: sure, jsut updated

sukhpalp (approved):

PR #2089, opened 2024-10-23
wfnews-2423

yzlucas:
Change desired_count for wfone_notificatiion_api ecs service to 1, upate version number. Update INSTANCE_COUNT to 1 in github environment variables.

### DIFF
--- a/terraform/ecs.tf
+++ b/terraform/ecs.tf
@@ -1500,7 +1500,7 @@ resource "aws_ecs_service" "wfone_notifications_api" {
   name                              = "wfone-notifications-api-${var.target_env}"
   cluster                           = aws_ecs_cluster.wfnews_main.id
   task_definition                   = aws_ecs_task_definition.wfone_notifications_api.arn
-  desired_count                     = var.app_count
+  desired_count                     = 1
   enable_ecs_managed_tags           = true
   propagate_tags                    = "TASK_DEFINITION"
   health_check_grace_period_seconds = 60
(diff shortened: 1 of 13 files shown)


terraform/ecs.tf
```diff
@@ -1500,7 +1500,7 @@ resource "aws_ecs_service" "wfone_notifications_api" {
   name                              = "wfone-notifications-api-${var.target_env}"
   cluster                           = aws_ecs_cluster.wfnews_main.id
   task_definition                   = aws_ecs_task_definition.wfone_notifications_api.arn
-  desired_count                     = var.app_count
```
sukhpalp, line 1503: Why are we making this a constant if we are updating the variable too?

terraform/variables.tf
```diff
@@ -88,7 +88,7 @@ variable "api_key" {
 
 variable "app_count" {
   description = "Number of docker containers to run"
-  default     = 2
+  default     = 1
```
sukhpalp, line 96: Did we confirm this with @vivid-cpreston ?

yzlucas: Yes, @vivid-cpreston also suggests to modify desired_count for ecs to use '1' to prevent issue from recuring

ssylver93, line 96: I think a new more specific variable like "notifications_api_app_count" that is set in Github actions and deployed in terraform should be used here in the case that we want to change the number of instances in future, to avoid another code change

ssylver93 (changes requested):
I think a new variable for notifications api instances should be used rather than a constant, as explained below

terraform/variables.tf
yzlucas: @ssylver93  sounds good! NOTIFICATIONS_API_INSTANCE_COUNT added in terrafrom and github

ssylver93 (approved):

PR #2091, opened 2024-10-24
wfmews-2425. handle url share between mobile and desktop 

yzlucas:
https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2425
Add URL query parameter transformation logic for mobile and desktop routes.
Handle query parameter transformations, purpose of these is because or mobile detail page for evac, area-restrictions, bans, dang rating are different as desktop detail page and uses different parameters. Eventually we might want to align them but it way too out of scope of WFNEWS-2425 and WFNEWS-2423

ie: desktop area-restriction detail page is: /events?eventType=area-restriction&eventNumber=2&eventName=C55175%20Area%20Restriction
but mobile area-restriction detail page is :/full-details?type=area-restriction&id=2&source=list&name=C55175%20Area%20Restriction

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/services/device-redirect-guard.ts
+++ b/client/wfnews-war/src/main/angular/src/app/services/device-redirect-guard.ts
@@ -1,11 +1,13 @@
 import { Injectable } from '@angular/core';
 import { CanActivate, Router, ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { isMobileView } from '@app/utils';
 import { Observable } from 'rxjs';
 
 @Injectable({
   providedIn: 'root'
 })
 export class DeviceRedirectGuard implements CanActivate {
+  public isMobileView = isMobileView;
 
   constructor(private router: Router) {}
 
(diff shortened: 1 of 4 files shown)


client/wfnews-war/src/main/angular/src/app/services/device-redirect-guard.ts
```diff
@@ -0,0 +1,106 @@
...
+})
+export class DeviceRedirectGuard implements CanActivate {
+
+  constructor(private router: Router) {}
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    const isMobile = (window.innerWidth < 768 && window.innerHeight < 1024) || (window.innerWidth < 1024 && window.innerHeight < 768);
```
sukhpalp: Instead of defining this here we should call an already existing service function or create one so these magic numbers do not fall out of sync

client/wfnews-war/src/main/angular/src/app/services/device-redirect-guard.ts
```diff
@@ -0,0 +1,106 @@
...
+    // If conditions are met, allow the route to activate
+    return true;
+  }
+
+  // Function to handle query parameter transformations
+  private transformQueryParams(queryParams: URLSearchParams): void {
+    const eventType = queryParams.get('eventType');
+    
+    switch (eventType) {
+      case 'Order':
```
sukhpalp: Do we have constants for these?

sukhpalp (changes requested):
Some changes around keeping the code DRY

yzlucas:
> Some changes around keeping the code DRY

@sukhpalp Done, also created constanst for Type and EventType to make the code clean and esay to maintain

sukhpalp (approved):
👍

PR #2094, opened 2024-10-25
wfnews-2418

yzlucas:
https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2418
Move the order of ROF steps. (move call back step to the end) 
Text changes based on mock up https://www.figma.com/design/ycmiyj65cNu5z1ywOj5BWq/WFNEWS?node-id=4253-27347&node-type=text&m=dev
image

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/contact-page/rof-contact-page.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/contact-page/rof-contact-page.component.ts
@@ -59,11 +59,26 @@ export class RoFContactPage extends RoFPage {
     }
   }
 
+  get iconType(): string {
+    return this.isFormValid ? 'arrow-forward-enabled' : 'arrow-forward-disabled';
+  }
+
+  get isFormValid(): boolean {
+    return !!this.reportOfFire.fullName && 
+          !!this.reportOfFire.phoneNumber && 
+          this.validatePhoneNumber(this.reportOfFire.phoneNumber);
+  }
+
+  get buttonClass(): string {
+    return this.isFormValid ? 'rof-button-primary' : 'rof-button-disabled';
+  }
+
   nextPage() {
     if (this.reportOfFire.motionSensor !== 'no' && !this.commonUtilityService.checkIfLandscapeMode()) {
       this.next();
     } else {
-      this.reportOfFirePage.selectPage('distance-page', null, false);
+      this.reportOfFirePage.selectPage('review-page', null, false);
     }
   }
+  
 }
--- a/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/permissions-page/rof-permissions-page.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/report-of-fire/permissions-page/rof-permissions-page.component.ts
@@ -1,6 +1,8 @@
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { RoFPage } from '../rofPage';
 import { ReportOfFire } from '../reportOfFireModel';
+import { ReportOfFirePage } from '@app/components/report-of-fire/report-of-fire.component';
+import { CommonUtilityService } from '@app/services/common-utility.service';
 
 @Component({
   selector: 'rof-permissions-page',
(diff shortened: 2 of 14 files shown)


client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.ts
```diff
@@ -121,10 +126,42 @@ this.reportOfFire.headingDetectionActive = true;
...
+
+    getButtonText(): string {
+      if (this.showProgress || this.id === 'callback-page' || this.id === 'contact-page') {
+        return 'Next';
+      }
+      return 'Continue';
+    }
+  
+    handleConfirmButtonClick(): void {
+      if (this.id === 'callback-page' || this.id === 'contact-page') {
```
agordon-vivid, line 144: This seems like code that should be handled at the page level given you have page specific logic here.  I'd be either looking to inject this in from the page -> component or give the component a way to call up to the page.

It would save you all this nested conditional logic.  What do you think?

yzlucas: The fact is there are multiple pages in ROF share the same component and call-back page does not has his own component (ie: all the single select page uses 'rof-simple-question-page.component' )
I would suggest to keep it within the component for now to maintain the current structure.
Ideally I want to re-structure the whole ROF components to let every single page has their own component to maintain them eaiser, but might be too out of scope at this moment...

agordon-vivid: Upon further look at this work, while we want to have page agnostic components, now is not the time to make that change.  The logic is too intertwined now.

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.ts
```diff
@@ -45,6 +49,7 @@ export class RoFSimpleQuestionPage extends RoFPage {
   }
 
   processToNext() {
+    console.log("WTF WHERE AM I")
```
sukhpalp, line 52: Please clean up.

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/contact-page/rof-contact-page.component.ts
```diff
@@ -59,11 +59,26 @@ export class RoFContactPage extends RoFPage {
     }
   }
 
+  get iconType(): string {
+    return this.isFormValid ? 'arrow-forward-enabled' : 'arrow-forward-disabled';
+  }
+
+  get isFormValid(): boolean {
+    return !!this.reportOfFire.fullName && 
+          !!this.reportOfFire.phoneNumber &&
```
sukhpalp, line 68: if we are validating the phone number in the function below, we can probably perform `!!this.reportOfFire.phoneNumber` inside of that function.

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/permissions-page/rof-permissions-page.component.ts
```diff
@@ -22,4 +27,14 @@ export class RoFPermissionsPage extends RoFPage {
   dataShareAcceptedToggle(event: boolean) {
     this.dataShareAccepted = event;
   }
+
+  nextPage() {
+    if (this.reportOfFire.motionSensor === 'yes' && !this.commonUtilityService.checkIfLandscapeMode()) {
```
sukhpalp, line 32: Where is motionSensor sourced from? Can we make it boolean? If not we should have a function to do this check in the reportOfFire object.

agordon-vivid (approved):

sukhpalp (changes requested):
A few small areas for clean up.

client/wfnews-war/src/main/angular/src/app/components/report-of-fire/simple-question-page/rof-simple-question-page.component.ts
```diff
@@ -121,10 +126,42 @@ this.reportOfFire.headingDetectionActive = true;
...
       } else {
-        this.reportOfFirePage.selectPage('distance-page', null, false);
+        this.reportOfFirePage.selectPage('review-page', null, false);
       }
     } else {
       this.skip();
     }
   }
+
+    getButtonText(): string {
```
sukhpalp, line 136: I think this should be in the JSON instead of determined in code. It can be an optional property along the lines of buttonText which would take priority if populated.

sukhpalp (commented):
One more comment

PR #2096, opened 2024-10-28
WFNEWS-2411 Use CapacitorHttp on mobile to avoid CORS issues in highlights widget

ssylver93:
The standalone iOS app is not returning any results from the WordPress API due to CORS errors when using Angular's HttpClient. This change will detect if the Capacitor native platform is being used and fetch the blog posts for the highlights widget using CapacitorHttp if so.


yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
```diff
@@ -139,4 +141,22 @@ export class HighlightsWidgetComponent implements OnInit {
       fireCentres: fireCentreTags.map(tag => tag.name)  // Array of fire centre names
     };
   }
+
+  // Use CapacitorHttp if on mobile to avoid CORS issues
+  private processRequest<T>(url: string): Observable<T> {   
+    if (Capacitor.isNativePlatform()) {
```
sukhpalp: ~Is there a reason to not use CapacitorHttp? Could we not use it on non native as well?~

I misunderstood how this library worked. Please disregard this comment. The one below still applies though, and I think we should keep this logic in some sort of HttpService.

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
```diff
@@ -139,4 +141,22 @@ export class HighlightsWidgetComponent implements OnInit {
       fireCentres: fireCentreTags.map(tag => tag.name)  // Array of fire centre names
     };
   }
+
+  // Use CapacitorHttp if on mobile to avoid CORS issues
+  private processRequest<T>(url: string): Observable<T> {
```
agordon-vivid: I assume we're using capacitorhttp elsewhere already?  

Is this function something that should live in a more general context?  Will other components need to use this same technique?

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
sukhpalp: If we cannot use the CapacitorHttp for non native HTTP call, I would suggest that we write a wrapping service that switches between the two HTTP providers in a central service, as opposed to spreading that logic throughout the code.

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
agordon-vivid: The more I think about this, the more I'd like to see logic like this in a service class.

agordon-vivid (changes requested):
Plese move request logic to a service class.

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
ssylver93: Thanks, I moved the getRequest function to the common utility service now, and added some formatting/linting changes in the files I was working on.

client/wfnews-war/src/main/angular/src/app/components/dashboard-component/widgets/highlights-widget/highlights-widget.component.ts
ssylver93: Thanks, I moved the getRequest function to the common utility service now, and added some formatting/linting changes in the files I was working on.

agordon-vivid (approved):

sukhpalp (approved):

PR #2106, opened 2024-10-29
wfnews-2407

yzlucas:
The PDT/PST suffix on a timestamp on the full details page needs to automatically adjust depending on the time of the year
https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2407
image

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/utils/index.ts
+++ b/client/wfnews-war/src/main/angular/src/app/utils/index.ts
@@ -598,8 +598,8 @@ export function getStageOfControlIcon(code: string) {
   }
 }
 
-export function convertToDateTimeTimeZone(date) {
-  // e.g. July 19, 2022 at 10:22 a.m. PST
+export function convertToDateTimeTimeZone(date: Date | string): string {
+  // e.g. July 19, 2022 at 10:22 a.m. PST/PDT
   const updateOptions: Intl.DateTimeFormatOptions = {
     year: 'numeric',
     month: 'long',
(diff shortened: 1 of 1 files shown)


client/wfnews-war/src/main/angular/src/app/utils/index.ts
```diff
@@ -598,20 +598,23 @@ export function getStageOfControlIcon(code: string) {
...
+export function convertToDateTimeTimeZone(date: Date | string): string {
+  // e.g. July 19, 2022 at 10:22 a.m. PST/PDT
   const updateOptions: Intl.DateTimeFormatOptions = {
     year: 'numeric',
     month: 'long',
     day: 'numeric',
     hour: 'numeric',
     minute: 'numeric',
     second: undefined, // this removes the seconds
+    timeZone: 'America/Los_Angeles', // Pacific Time Zone
```
sukhpalp: Do we need to specify timezone? If somebody was using this from the eastern parts of BC, they would be getting the wrong time zone as they are one hour ahead. Would it be enough just to get rid of the fixed PST?

sukhpalp (changes requested):
I don't think we want to specify a timezone. If we use toLocaleString, it will base it off the user's client time.

client/wfnews-war/src/main/angular/src/app/utils/index.ts
yzlucas: You are right, I forgot to consider some part of BC are off one hour. 
Removed timezone option, thanks

sukhpalp (approved):

PR #2112, opened 2024-10-30
WFNEWS-2409 Change Area Restriction radius to 0.5KM

ssylver93:
Changing Area Restriction radius from 25KM to 0.5KM so only overlapping Area Restrictions are returned for incidents. Changing the class name from PublicIncidentPage to PublicIncidentPageComponent according to _(https://angular.io/styleguide#style-02-03)eslint[@angular-eslint/component-class-suffix] Component class names should end with one of these suffixes: "Component"_


client/wfnews-war/src/main/angular/src/app/app.module.ts
```diff
@@ -251,7 +251,7 @@ import { SuspectedCauseCardComponent } from './components/public-incident-page/i
 import { TextCardComponent } from './components/public-incident-page/incident-tabs/cards/text-card/text-card.component';
 import { ContactUsBannerComponent } from './components/public-incident-page/incident-tabs/contact-us-banner/contact-us-banner.component';
 import { IncidentTabsComponent } from './components/public-incident-page/incident-tabs/incident-tabs.component';
-import { PublicIncidentPage } from './components/public-incident-page/public-incident-page.component';
+import { PublicIncidentPageComponent } from './components/public-incident-page/public-incident-page.component';
```
sukhpalp, line 254: Thank you for cleaning up the SonarLint stuff as you go. 👍

sukhpalp (approved):

PR #2119, opened 2024-11-01
text and icon change for saved page on mobile

yzlucas:
image
Change the header, label and adding icons for the Save page. Had to add a new svg icon because the existed one has different size, color and resolution.

https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2415


sukhpalp (approved):
This works. Going forward we should try to reduce the number of icons being added and use masking with the existing icon. I think it would be nice to remove all of the icon color variants in favor of masking.

PR #2120, opened 2024-11-01
adding a photo on capacitor share

yzlucas:
https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2422
Add a generic image on share via mobile device


sukhpalp:
Don't forget to include the ticket number to the title or description.

sukhpalp (changes requested):
This ticket should also cover the OpenGraph meta tags, which allows apps to fetch the image when sharing the link outside of the Capacitor share.

https://www.tektutorialshub.com/angular/dynamic-meta-tags-in-angular/

yzlucas:
> This ticket should also cover the OpenGraph meta tags, which allows apps to fetch the image when sharing the link outside of the Capacitor share.
> 
> https://www.tektutorialshub.com/angular/dynamic-meta-tags-in-angular/

I added metaTags on sharing feature, please review. Seems like Twitter uses diffrent property for tags, Facebook and other platforms uses openGraph

sukhpalp (approved):

yzlucas:
update og tage and twitter tag based on Christine's comment in the ticket https://apps.nrs.gov.bc.ca/int/jira/browse/WFNEWS-2422

yzlucas:
added `    this.router.events
    .pipe(filter(event => event instanceof NavigationEnd))
    .subscribe(() => {
      this.setDefaultMetaTags();
    });`
    
    in app.component.ts to reset the default meta tags on page changing.

PR #2139, opened 2024-11-07
add build number injection in ios and android pipeline

yzlucas:


thefunkyhermit (approved):

sukhpalp (changes requested):
~I think it would be cleaner to add a property to appConfig.mobile.json and update the job to replace the token in the file at build time.~

On second thought, this is fine. If we are using this file on desktop too, then this is consistent.

ssylver93 (approved):

sukhpalp (approved):

PR #2151, opened 2024-12-14
WFNEWS-2501: typo of snowplow action name, also fix the WFNEWS-2514: associate wildfire navigate does not work on area-restriction page

yzlucas:


client/wfnews-war/src/main/angular/src/app/components/full-details/area-restrictions-full-details/area-restrictions-full-details.component.ts
```diff
@@ -1,5 +1,5 @@
 import { ChangeDetectorRef, Component, Input, OnInit } from '@angular/core';
-import { Router as Route } from '@angular/router';
+import { Router as Route, Router } from '@angular/router';
```
ssylver93: This looks like a double import to me? Router is already imported

ssylver93 (changes requested):
I think Router is getting imported twice

agordon-vivid (approved):

client/wfnews-war/src/main/angular/src/app/components/full-details/area-restrictions-full-details/area-restrictions-full-details.component.ts
yzlucas: good catch

PR #2153, opened 2024-12-17
WFNEWS-2513 Text changes for response type description and title

ssylver93:
Text changes for response type description
Change title to "Response"
Remove secondary title

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/utils/index.ts
+++ b/client/wfnews-war/src/main/angular/src/app/utils/index.ts
@@ -954,5 +954,3 @@ export function zoomInWithLocationPin(){
       ).addTo(viewer.map);
     }
   }
-
-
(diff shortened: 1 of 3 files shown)


yzlucas (approved):

client/wfnews-war/src/main/angular/src/app/utils/index.ts
```diff
@@ -806,7 +806,9 @@ export function formatDate(timestamp: string | number): string {
     };
 
     return date.toLocaleDateString('en-US', options);
-  } else return '';
```
agordon-vivid, line 809: should we throw an exception instead of a blank string?

agordon-vivid (approved):
Non-blocking question

client/wfnews-war/src/main/angular/src/app/utils/index.ts
ssylver93: Yes, updated now

PR #2157, opened 2024-12-17
fix the method for tab to pass value to snowplow

yzlucas:
When using a custom template for labels (<ng-template mat-tab-label>), Angular Material does not populate the tab.textLabel property. Therefore, tab.textLabel is often undefined or an empty string.

### DIFF
--- a/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/public-incident-page.component.ts
+++ b/client/wfnews-war/src/main/angular/src/app/components/public-incident-page/public-incident-page.component.ts
@@ -294,19 +294,27 @@ export class PublicIncidentPageComponent implements OnInit {
     window.location.href = mailtoUrl;
   }
 
-  onTabChange(event: MatTabChangeEvent) {
+  onTabChange(event: MatTabChangeEvent | number): void {
+    const tabLabels = ['Details', 'Response', 'Gallery', 'Maps'];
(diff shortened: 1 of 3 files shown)


client/wfnews-war/src/main/angular/src/app/components/public-incident-page/public-incident-page.component.ts
```diff
@@ -294,19 +294,27 @@ export class PublicIncidentPageComponent implements OnInit {
...
+  onTabChange(event: MatTabChangeEvent | number): void {
+    const tabLabels = ['Details', 'Response', 'Gallery', 'Maps'];
     const url = this.appConfigService.getConfig().application.baseUrl.toString() + this.currentRouter.url.slice(1);
+  
+    // Determine the index based on the type of event
+    const index = typeof event === 'number' ? event : event.index;
+  
     let actionName;
-    if (event?.tab?.textLabel === 'Response') {
+    if (index === 0) {
```
agordon-vivid: For readability, I'd set the numbers you're checking to ENUMs.  But it's not a blocker on this PR.

agordon-vivid (approved):

client/wfnews-war/src/main/angular/src/app/components/public-incident-page/public-incident-page.component.ts
yzlucas: Agree, refactored a bit by creating a TAB_ACTIONS array for readbility

PR #2195, opened 2025-12-12
Update environment names in URLs

vivid-cpreston:


.github/workflows/android.yml
```diff
@@ -56,6 +56,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - shell: bash
+        id: map-environment
```
sukhpalp, line 60: Do we need one for the dlv environment?

sukhpalp (approved):

.github/workflows/android.yml
vivid-cpreston: We're not planning on adding a DLVR environment for WFNEWS, but I've added support of it to the map-environment bit in case that changes

PR #2214, opened 2026-02-17
ROF scheduled delete

sukhpalp:
Instead of deleting inline, this deletes the old ROFs after 48 hours. This would allow for duplicate checking to work for ROFs where the retry were spaced out.

Disable submit button on submit.


vivid-cpreston (approved):
Concept sounds great! Code looks good to me, but you'll want someone more familiar with this for a proper code review

yzlucas (approved):
Approved! The backend change makes the create operation safe against duplicate requests.

PR #2225, opened 2026-05-06
Feature/wfnews 2761 import resources

vivid-cpreston:


sukhpalp:
Do we need to add permissions to these workflows?

sukhpalp (approved):
I think you may need to add the permission blocks in order for these to work

sukhpalp (approved):

